PR #11, opened 2020-12-28
Animation update

petrie911:
Should work with the new structs. Experimented with alternate formatting for the frame data, but it didn't work for longer animations.

### DIFF
--- a/ZAP2/ZAnimation.cpp
+++ b/ZAP2/ZAnimation.cpp
@@ -86,8 +86,8 @@ string ZAnimation::GetSourceOutputCode(string prefix)
 
 		string indicesStr = "";
 		string valuesStr = "\t";
-		int lineLength = 15;
-		int offset = 0;
+		const int lineLength = 15;
+		const int offset = 0;
 
 		for (int i = 0; i < rotationValues.size(); i++)
 		{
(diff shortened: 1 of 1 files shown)


ZAP2/ZAnimation.cpp
```diff
@@ -80,28 +80,30 @@ string ZAnimation::GetSourceOutputCode(string prefix)
...
 	{
-		string headerStr = StringHelper::Sprintf("\t { %i, %i }, %s_values, %s_indices, %i",
-			frameCount, 0, name.c_str(), name.c_str(), limit);
+		string headerStr = StringHelper::Sprintf("%i, %s_values, %s_indices, %i",
+			frameCount, name.c_str(), name.c_str(), limit);
 		parent->declarations[rawDataIndex] = new Declaration(DeclarationAlignment::None, 16, "AnimationHeader", StringHelper::Sprintf("%s", name.c_str()), false, headerStr);
 
 		string indicesStr = "";
 		string valuesStr = "\t";
+		int lineLength = 15;
```
NEstelami: Minor thing but if these two lines aren't going to change, please make them constant.

petrie911: Done.

PR #39, opened 2021-01-01
No longer using snake case name variables and segment offsets now print to 6 characters

Kenix3:


ZAPD/ZDisplayList.cpp
```diff
@@ -182,14 +182,14 @@ int ZDisplayList::OptimizationCheck_LoadTextureBlock(int startIndex, string& out
 				if (texDecl != nullptr)
 					texStr = StringHelper::Sprintf("%s", texDecl->varName.c_str());
 				else if (segmentNumber == 2)
-					texStr = StringHelper::Sprintf("%s_tex_%08X", scene->GetName().c_str(), texAddr);
+					texStr = StringHelper::Sprintf("%sTex0x%06X", scene->GetName().c_str(), texAddr);
```
fig02, line 185: I was going to leave an underscore for offsets, similar to how we do `unk_123` with variables in code. since the segment offsets are only really being printed for unnamed stuff. 
May even want to omit the 0x like we do with unks, but want to hear other opinions

fig02: we can decide on this later tho no reason for it to hold things up. and like i mentioned in discord this should only be relevant for undefined names

PR #47, opened 2021-01-04
Fix <filesystem> check

leoetlino:
GCC >= 9 supports `<filesystem>` and doesn't require the use of
`<experimental/filesystem>` (which would require linking libstdc++fs...)


Kenix3 (approved):
Looks good to me.

PR #49, opened 2021-01-04
Bitconverter now supports more types and should correctly read floating point values.

Kenix3:


ZAPD/BitConverter.h
```diff
@@ -6,38 +6,107 @@
...
+	static inline uint64_t ToUInt64BE(const std::vector<uint8_t>& data, int offset)
 	{
 		return ((uint64_t)data[offset + 0] << 56) + ((uint64_t)data[offset + 1] << 48) + ((uint64_t)data[offset + 2] << 40) + ((uint64_t)data[offset + 3] << 32) + ((uint64_t)data[offset + 4] << 24) + ((uint64_t)data[offset + 5] << 16) + ((uint64_t)data[offset + 6] << 8) + ((uint64_t)data[offset + 7]);
 	}
 
-	static inline float ToFloatBE(std::vector<uint8_t> data, int offset)
+	static inline float ToFloatBE(const uint8_t* data, int offset)
+	{
+		uint32_t floatData = ((uint32_t)data[offset + 0] << 24) + ((uint32_t)data[offset + 1] << 16) + ((uint32_t)data[offset + 2] << 8) + (uint32_t)data[offset + 3];
+		return *(float*)(&floatData);
```
leoetlino: This kind of type punning is undefined behavior in standard C++. I suggest doing this instead:

```cpp
float value;
static_assert(sizeof(uint32_t) == sizeof(float));
std::memcpy(&value, &floatData, sizeof(value));  // include <cstring> for std::memcpy
return value;
```

The compiler will optimize away the call and you also get rid of the UB.

(If this project targeted C++20, you could use [std::bit_cast](https://en.cppreference.com/w/cpp/numeric/bit_cast).)

Kenix3: Good idea.
I should add that the byteswapping performed in all of the data extraction methods in that class are all implementation defined since ZAPD is not guaranteed to be running on a little endian system.

PR #50, opened 2021-01-04
ZAPD now supports multiple verbosity levels

Kenix3:
This wastes CPU cycles and creates unnecessary IO time. ZAPD doesn't need to output this.

### DIFF
--- a/ZAPD/Globals.cpp
+++ b/ZAPD/Globals.cpp
@@ -19,11 +19,11 @@ Globals::Globals()
 	segmentRefFiles = map<int, ZFile*>();
 	genSourceFile = true;
 	testMode = false;
-	debugMessages = false;
 	profile = false;
 	includeFilePrefix = false;
 	useExternalResources = true;
 	lastScene = nullptr;
+	verbosity = VERBOSITY_SILENT;
 }
 
 string Globals::FindSymbolSegRef(int segNumber, uint32_t symbolAddress)
--- a/ZAPD/Globals.h
+++ b/ZAPD/Globals.h
@@ -17,6 +17,7 @@ public:
 	bool debugMessages; // Enables certain printfs
 	bool profile; // Measure performance of certain operations
 	bool includeFilePrefix; // Include the file prefix in symbols
+	bool verbose = false; // ZAPD outputs additional information
 	ZFileMode fileMode;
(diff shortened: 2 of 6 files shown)


NEstelami:
> This wastes CPU cycles and creates unnecessary IO time. ZAPD doesn't need to output this.

I'm not against removing it, but do we know for certain this is actually taking up a noticeable amount of time? I hope we're not micro-optimizing something that is ultimately negligible here.

Kenix3:
> 
> 
> > This wastes CPU cycles and creates unnecessary IO time. ZAPD doesn't need to output this.
> 
> I'm not against removing it, but do we know for certain this is actually taking up a noticeable amount of time? I hope we're not micro-optimizing something that is ultimately negligible here.

I didn't benchmark it, but I can say that it crowds the output console with messages that have no use in all contexts.

louist103:
Personally I like having it have this because it makes it easier to see where it starts a new file.

NEstelami:
> Personally I like having it have this because it makes it easier to see where it starts a new file.

Maybe we should have a silent mode then?

Kenix3:
> 
> 
> > Personally I like having it have this because it makes it easier to see where it starts a new file.
> 
> Maybe we should have a silent mode then?

General behaviour of binaries is that silent mode is default and verbose is a passed in flag. I think it's a great compromise to add a verbose flag.

NEstelami:
> > > Personally I like having it have this because it makes it easier to see where it starts a new file.
> > 
> > 
> > Maybe we should have a silent mode then?
> 
> General behaviour of binaries is that silent mode is default and verbose is a passed in flag. I think it's a great compromise to add a verbose flag.

Sounds good to me. `-v` then?

Kenix3:
> 
> 
> > > > Personally I like having it have this because it makes it easier to see where it starts a new file.
> > > 
> > > 
> > > Maybe we should have a silent mode then?
> > 
> > 
> > General behaviour of binaries is that silent mode is default and verbose is a passed in flag. I think it's a great compromise to add a verbose flag.
> 
> Sounds good to me. `-v` then?

This message now relies on the -v flag. NOTE that there is an existing debugMessages flag. Should we roll all of the output for that flag into -v, or keep them separate for now? If we keep them separate then the -v flag only controls the intro message.

NEstelami (approved):
LGTM

PR #51, opened 2021-01-05
Jenkins

ethteck:
The repo now gets tested on Jenkins and built against OOT to make sure no changes break the build


Kenix3 (approved):
Looks good to me! It'll be nice to no longer have to manually build OOT.

PR #54, opened 2021-01-05
ZAPD now supports scalar types

Kenix3:
Additionally wrapped a few output statements in verbosity checks.
Additionally added checks for environment float types in BitConverter

### DIFF
--- a/ZAPD/ZScalar.cpp
+++ b/ZAPD/ZScalar.cpp
@@ -1,0 +1,219 @@
+#include "ZScalar.h"
+#include "ZFile.h"
+#include "BitConverter.h"
+#include "StringHelper.h"
+#include "File.h"
+#include "HighLevel/HLAnimationIntermediette.h"
+#include "Globals.h"
+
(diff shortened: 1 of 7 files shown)


ZAPD/ZScalar.cpp
```diff
@@ -0,0 +1,224 @@
...
+	default:
+		return "\"BADTYPE!!\"";
+	}
+}
+
+std::string ZScalar::GetSourceOutputCode(const std::string& prefix)
+{
+	if (parent != nullptr)
+	{
+		parent->declarations[rawDataIndex] = new Declaration(DeclarationAlignment::None, 16, GetSourceTypeName(), name, false, GetSourceValue());
```
NEstelami: Rather than writing to parent->declarations[] directly, please use one of the functions, such as parent->AddDeclaration(...). Accessing the dictionary directly is being depricated.

NEstelami (approved):
LGTM

PR #55, opened 2021-01-05
Made a bunch of declarations static. Removed external writes to declarations dictionary.

NEstelami:
Static declarations are prototyped at the top of the C file now. Some of these may be redundant, but until we add in a system to detect if a prototype is needed based on declaration order, it'll have to do. Overall with the number of vtx declarations, the header files should be a fair bit slimmer now due to not having a bunch of externs anymore.


Kenix3 (approved):
Looks good to me.

PR #63, opened 2021-01-06
Add a clang-format config and reformat all source files

leoetlino:
The existing code style is mostly preserved; this just makes formatting
more consistent and prevents lines from getting too long.


fig02:
Think we are going to revisit this, so if you can rebase whenever you have a sec leo, tyvm

fig02 (approved):

Issue #52, opened 2021-01-05
ZAP allocates a lot of memory and resources that are never freed.


Kenix3:
This is not meant to be a 100% exhaustive list.
- [x] XML handle may not be closed.
- [ ] File handles may not be closed.
- [x] Strings that are not stack variables need to be freed. Some come from XML, and it may handle freeing of those. We need to consult documentation.
- [x] ZFile.resources needs to free and destroy the ZResource on ZFile destructor.
- [x] ZFiles need to be destroyed after the XML file is done being processed. They are allocated memory in a few places, but never freed.
- [x] ZCollision allocates VertexEntry classes into the vertices vector and never frees them.
- [x] ZCollision allocates PolygonEntry classes into the polygons vector and never frees them.
- [x] ZCollision allocates a CameraDataList and never frees it.
- [x] ZCollision allocates WaterBoxHeader classes into the waterBoxes vector and never frees them.
- [x] ZCutscene allocates commands into the commands vector and never frees them.
- [ ] ZDisplayList allocates ZDisplayLists into the otherDLists vector and never frees them.
- [x] File.ReadAllBytes allocates a char array that is never freed.
- [x] File.ReadAllText allocates a char array that is never freed.
- [x] SetListActor allocates ActorSpawnEntry to the actors vector and never frees them.
- [x] SetCutscenes allocates ZCutscene to the cutscenes vector and never frees them.
- [x] SetEntranceList allocates EntranceEntry to the entrances vector and never frees them.
- [x] SetLightSettings allocates LightingSettings to the settings vectors and never frees them.
- [x] SetMesh allocates MeshHeader and other objects, but never frees them.
- [x] SetPathways allocates PathwayEntry to the pathways vector and never frees them.
- [x] SetRoomList allocates RoomEntry to the rooms vector and never frees them.
- [x] SetStartPositionList allocates ActorSpawnEntry to the actors vector and never frees them.
- [x] SetTransitionActorList allocates TransitionActorEntry to the transitionActors vector and never frees them.
- [ ] ZRoom.ExtractFromXML allocates multiple objects and never frees them.
- [x] ZRoom allocates room commands to the commandList vector and never frees them.
- [x] ZLimbStandard.FromRawData allocates a ZLimbStandard that is never freed or added to the ZResource list in a ZFile.
- [x] ZLimbLOD::FromRawData allocates a ZLimbLOD that is never freed or added to the ZResource list in a ZFile.
[...]

PR #83, opened 2021-02-14
Fix some memory leaks issues

AngheloAlf:
I mainly changed a lot of raw pointers into `shared_ptr`, so that object deallocates itself when it is no longer used. I tried finding the places where the memory was leaking and adding `delete`s, but they were too much places and also it somehow randomly crashed during my testing, so I just went with the `shared_ptr` solution.

There were also a lot of places where a heap allocation (`new`) was used when it wasn't necessary, and for some reason the object wasn't being `delete`d, so it just leaked...

I would suggest to avoid using pointers unless it is really neaded (like storing a bunch of objects of different types using inheritance). And even in that case, try to use smart pointers so you don't have to worry about deallocating that memory.

This should fix almost all the problems listed in #52 (except the last one).

And for some numbers, this should fix around 50% of leaked memory. 
To get this number, I measured the leaked memory when running `extract_assets.py` with and without this change. It went down from `1254.07 MiB` leaked to `619.66 MiB` leaked.

There are still things to fix in `HighLevel` and `ZRoom` folders. I tried to fix stuff there, but it exploded in my face... Maybe I'll try to fix them later.

### DIFF
--- a/ZAPD/File.h
+++ b/ZAPD/File.h
@@ -14,7 +14,8 @@ public:
 	static bool Exists(const std::string& filePath)
 	{
 		std::ifstream file(filePath, std::ios::in | std::ios::binary | std::ios::ate);
-		return file.good();
+		bool good = file.good();
+		return good;
 	}
 
 	static std::vector<uint8_t> ReadAllBytes(const std::string& filePath)
@@ -26,6 +27,7 @@ public:
 		file.read(data, fileSize);
 		std::vector<uint8_t> result = std::vector<uint8_t>(data, data + fileSize);
 		delete[] data;
+		file.close();
 
 		return result;
 	};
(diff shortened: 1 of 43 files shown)


NEstelami:
Don't worry about the stuff in the `HighLevel` folder. It's all experimental and doesn't really apply to the extraction process.

AngheloAlf:
> 
> 
> Don't worry about the stuff in the `HighLevel` folder. It's all experimental and doesn't really apply to the extraction process.

The changes I made in `HighLevel` are minimal. I made them because ZAPD will not compile otherwise.

fig02:
theres a handful of conflicts to resolve in here, unsure if we want to do this before or after clang-format. up to you alf

AngheloAlf:
I think I'll start over with this. There are many new changes, and now that I now a lot more of ZAPD I think I could do a better job starting over.

PR #84, opened 2021-02-27
Use libgfxd (gfxdis) for display lists

fig02:
Big thanks to glank and tharo for the help with this.

This set of changes uses the new gfxdis library to disassemble display list data.
The basic idea behind how this works is: 
Binary display list data is passed to the gfxdis library, and you can write callback functions to handle different commands a certain way. ZAP still handles the declarations as it always does. I ported the important parts of existing code to determine symbol names to the callbacks.

With this set of changes, alot of the code in ZDisplayList.cpp that does the actual disassembling is not needed anymore. But I figured I should get an OK before refactoring.

As is, this OKs with OoTs current master. That said, I dont think this is the most elegant way to implement this. We could maybe afford to refactor alot of how this works down the line. But this works for our purposes currently.

Also included is a new class OutputFormatter written by glank. Currently this is used to convert tabs to spaces and enforce a 120 character line limit on display lists. This can easily be extended to be used on all of the extraction output, which we should look into doing soon.


Zelllll (approved):
Looks good to me, this should be merged.

PR #85, opened 2021-03-04
Fix collision extraction from actor's data

AngheloAlf:
Fixes a crash when ZAPD tries to extract a CollisionHeader from overlay's data. 

The issue was caused because ZAPD didn't account the base address of the overlay.

The issue can be replicated creating an xml for the overlay `ovl_En_Jsjutan` and the following:
- `BaseAddress="0x80A897A0" RangeStart="0x12F8" RangeEnd="0x4C88" Segment="128"`
- `<Collision Name="D_80A8E400" Offset="0x4C60"/>`

### DIFF
--- a/ZAPD/ZAnimation.cpp
+++ b/ZAPD/ZAnimation.cpp
@@ -183,6 +183,6 @@ void ZLinkAnimation::ParseRawData()
 
 	const uint8_t* data = rawData.data();
 
-	//segmentAddress = SEG2FILESPACE(BitConverter::ToInt32BE(data, rawDataIndex + 4));
+	//segmentAddress = GETSEGOFFSET(BitConverter::ToInt32BE(data, rawDataIndex + 4));
 	segmentAddress = (BitConverter::ToInt32BE(data, rawDataIndex + 4));
 }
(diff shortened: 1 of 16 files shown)


mzxrules:
Most of the changes seem good. The main issue I have is that this pattern...

```c
	segmentNumber = GETSEGNUM(polySegmentOffset);
	if (segmentNumber == 0x80) { // Is defined in code?
		polyAddr -= SEG2FILESPACE(parent->baseAddress);
	}
```
...should ideally be part of SE2FILESPACE. The laziest/quickest thing to do would probably to just make it into it's own separate function

AngheloAlf:
I just followed the same pattern as in `ZDisplayList` (Example [1](https://github.com/zeldaret/ZAPD/blob/d9ba72302b1105d79a06c7da8d9c1bbb8ba2fceb/ZAPD/ZDisplayList.cpp#L801), [2](https://github.com/zeldaret/ZAPD/blob/d9ba72302b1105d79a06c7da8d9c1bbb8ba2fceb/ZAPD/ZDisplayList.cpp#L823) and [3](https://github.com/zeldaret/ZAPD/blob/d9ba72302b1105d79a06c7da8d9c1bbb8ba2fceb/ZAPD/ZDisplayList.cpp#L905)).

I'll try and add this check into `SEG2FILESPACE` macro.

AngheloAlf:
It should be fixed now.

PR #87, opened 2021-03-10
A few bugfixes

AngheloAlf:
Some little fixes that i didn't think should have their own PR.
- Remove an extra `%i` in `CS_SCENE_TRANS_FX(%i, %i, %i)`. 
  - This macro uses only 3 parameters, not four.
  - Also, ZAPD was writing junk to that fourth value.
- `ZDisplayList`: Don't extract vertices from a different segment declared in this file.
  - For example, if an object in segment 06 references a Vtx in segment 08, currently ZAPD will try to export that Vtx from the current file. With this change, ZAPD will just leave the addresses instead of symbols inside the dlist.
  - To test this behaviour, see the dlist in `object_fhg` with offset `0xCB30`.
  - ~The `gfxdis` (#84) pr has this same problem too~ Not anymore.
- ~`ZCollision`: Change the extraction from `polygonTypes` (`u32`) to `surfaceType` (`SurfaceType`) to match the current decomp repo~ Fixed by #84.


ZAPD/ZDisplayList.cpp
```diff
@@ -810,6 +810,13 @@ void ZDisplayList::Opcode_G_VTX(uint64_t data, int i, std::string prefix, char*
 		sprintf(line, "gsSPVertex(@r, %i, %i),", nn, _SHIFTR(hi, 17, 7));
 	}
 
+	// Hack: Don't extract vertices from a unknown segment.
+	if (!Globals::Instance->HasSegment(GETSEGNUM(data))) {
+		segptr_t segmented = data & 0xFFFFFFFF;
+		references.push_back(segmented);
+		parent->AddDeclaration(segmented, DeclarationAlignment::Align16, 16, "Vtx", StringHelper::Sprintf("0x%08X", segmented), "");
+		return;
+	}
```
fig02, line 821: I dont follow whats happening here, if the data is from an unknown segment how come you need to do references.pushback and add a declaration?

only asking cause this same fix needs to be implemented in the gfxdis branch

fig02: i implemented a different fix here: https://github.com/fig02/ZAPD/blob/gfxdis/ZAPD/ZDisplayList.cpp#L1453
If you find the time to try it out with object_fhg, let me know if it works

fig02 (approved):

PR #88, opened 2021-03-11
std::map is already a sorted container, no need to sort it

gamestabled:


ZAPD/ZFile.cpp
```diff
@@ -793,11 +787,11 @@ string ZFile::ProcessDeclarations()
 		ProcessDeclarationText(item.second);
 	}
 
-	for (pair<int32_t, Declaration*> item : declarationKeysSorted)
+	for (pair<int32_t, Declaration*> item : declarations)
```
AngheloAlf, line 810: Since you are changing this for performance reasons, you may want to use a reference in the iterator variable, so it doesn't perform a copy in each iteration.

```suggestion
	for (pair<int32_t, Declaration*>& item : declarations)
```

ZAPD/ZFile.cpp
```diff
@@ -877,7 +869,7 @@ string ZFile::ProcessDeclarations()
 	// Handle unaccounted data
 	lastAddr = 0;
 	lastSize = 0;
-	for (pair<int32_t, Declaration*> item : declarationKeysSorted)
+	for (pair<int32_t, Declaration*> item : declarations)
```
AngheloAlf, line 892: Same here
```suggestion
	for (pair<int32_t, Declaration*>& item : declarations)
```

ZAPD/ZFile.cpp
```diff
@@ -945,15 +937,9 @@ string ZFile::ProcessDeclarations()
...
-	declarationKeysSorted = vector<pair<int32_t, Declaration*>>(declarations.begin(), declarations.end());
-	sort(declarationKeysSorted.begin(), declarationKeysSorted.end(), [](const auto& lhs, const auto& rhs)
-	{
-		return lhs.first < rhs.first;
-	});
-
 	// First, handle the prototypes (static only for now)
 	int protoCnt = 0;
-	for (pair<int32_t, Declaration*> item : declarationKeysSorted)
+	for (pair<int32_t, Declaration*> item : declarations)
```
AngheloAlf, line 962: Same here
```suggestion
	for (pair<int32_t, Declaration*>& item : declarations)
```

ZAPD/ZFile.cpp
```diff
@@ -975,7 +961,7 @@ string ZFile::ProcessDeclarations()
 		output += "\n";
 
 	// Next, output the actual declarations
-	for (pair<int32_t, Declaration*> item : declarationKeysSorted)
+	for (pair<int32_t, Declaration*> item : declarations)
```
AngheloAlf, line 984: Same here
```suggestion
	for (pair<int32_t, Declaration*>& item : declarations)
```

ZAPD/ZFile.cpp
```diff
@@ -1072,13 +1058,7 @@ string ZFile::ProcessExterns()
...
 	string output = "";
 
-	auto declarationKeysSorted = vector<pair<int32_t, Declaration*>>(declarations.begin(), declarations.end());
-	sort(declarationKeysSorted.begin(), declarationKeysSorted.end(), [](const auto& lhs, const auto& rhs)
-	{
-		return lhs.first < rhs.first;
-	});
-
-	for (pair<int32_t, Declaration*> item : declarationKeysSorted)
+	for (pair<int32_t, Declaration*> item : declarations)
```
AngheloAlf, line 1085: Same here
```suggestion
	for (pair<int32_t, Declaration*>& item : declarations)
```

AngheloAlf (changes requested):
Nice one!
I like this change, it should improve performance at the cost of nothing.
I left you a few suggestions to improve it even more. 
Remember to test the OoT repo, it should still give an OK.

ethteck:
@AngheloAlf Jenkins actually tests the oot repo so no need :)

ZAPD/ZFile.cpp
simonlindholm: I'm not sure that actually helps performance -- the loop uses both item.first and item.second, so the copy is needed anyway, and if it wasn't the compiler would be able to optimize it out. On the other hand, using a reference may require a read for each use of .first/.second, depending on the compiler's ability to do alias analysis. (The concrete impact on perf is probably very small, though.) IME references start being useful only when there are non-trivial copy constructors involved, like if you have a `pair<std::string, ...>`.

gamestabled: My inclination (especially after simon's comment) is to leave it as is, but I don't feel strongly either way lol

NEstelami: I would keep it as-is. Even if there's an optimization there, it's going to be negligible either way.

AngheloAlf (approved):

fig02 (approved):

fig02 (approved):
conflict resolved

PR #96, opened 2021-03-18
Configure Jenkinsfile to build on new "slug" Jenkins agent

ethteck:
This is part of a larger effort to migrate all builds over to slug. The jenkinsfile now specifies a label requirement, so only agents matching the `ZAPD` label will attempt to build it.

I also removed the manual usage of ccache in the Makefile, as this is now being expected to be configured at the system level.


ethteck:
Build times on master agent:
image

Build times on new agent (slug):
image

ethteck:
Also thanks to @AngheloAlf for your PR fixing a build issue. I think we should wait for your PR to be merged before this one.

PR #95, opened 2021-03-18
Fix `make setup`

AngheloAlf:
Fixes the compilation problem when trying to compile ZAPD in the OoT repo by doing `make setup`.

PR #97, opened 2021-03-18
Fix repo warnings

AngheloAlf:
Fixes most of the warnings when compiling the oot repo.

~~The only one remaining to date should be the `Incompatible pointer type assignment` in `SkeletonHeader` and `FlexSkeletonHeader`.~~


fig02 (approved):
jenkins fails on this so we gotta figure that out, but everything else looks good

ethteck:
I think master just needs to be merged into it

AngheloAlf:
Actually, I was waiting for #95 to be merged. It should fix it.

PR #106, opened 2021-04-01
Add prerenders extraction

AngheloAlf:
Add extraction of prerenders used in some scenes like shops, houses, market, etc.
The prerenders will be extracted as a `.jfif` image (basically a jpeg).
Those can be recompiled back into an array.
There are 2 ways to export the prerenders:
- Automatically. If a scene declared in a xml has a prerender, it will be detected automatically and extracted with a autogenerated name.
- Declaring it manually in the xml. This is useful if wanted to provide a filename instead of using the autogenerated.
  - A prerender can be declared with two tags: `<Prerender>` (if declared inside a `<File>`) or `<PrerenderHint>` (if declared inside a `<Room>` or `<Scene>`).
  - Both tags expects a `Name`, `OutName` and `Offset` attributes.

A small change in the OoT repo makefile will have to be made to be able to compile the jfif prerenders into C code.
The change is attached as a patch. 
[0001-update-makefile.patch.txt](https://github.com/zeldaret/ZAPD/files/6246882/0001-update-makefile.patch.txt)

(I'll probably do a PR in OoT repo to fix the makefile, but it is added for completeness)


NEstelami:
Maybe it would make more sense for it to extract to `.jpg`? I understand the header says JFIF, but it's for all intents and purposes a JPEG file, and image editing/viewing software tend to refer to it as `.jpg` over `.jfif`. Modern `.jpg` files also have the `JFIF` magic in the header.

AngheloAlf:
If that the case, I'll change it.

AngheloAlf:
Oh, Jenkins is failing because it can't compile the OoT repo, because that repo doesn't have the changes needed in the makefile :c

Issue #16, opened 2020-12-29
ZAPD should be able to extract JPEG images.


Kenix3:
These are primarily used in pre-rendered rooms.
We should scrub any JFIF metadata before compiling back into a ROM.

PR #107, opened 2021-04-03
Rename "Prerender" to "Background", properly extract PolygonType1 multi format and change some structs used.

AngheloAlf:
- Renames "Prerender" to "Background"
- Extract PolygonType1 multi format.
  - Places like the "back alley" and "Temple of time" use this format for their prerenders/backgrounds.
- Use _some_ structs defined in `z64.h` instead of their versions of `z64scene.h`.
  - `PolygonDlist` instead of `MeshEntry0`.
  - `BgImage` instead of `BackgroundRecord`.
- Improve a bit the output code for scenes and rooms.


AngheloAlf:
Before I forget, merging this PR should close #16.

AngheloAlf:
I rebased master, so now this PR doesn't have any conflict left!

Also re added some stuff that to prevent warnings when building the OoT repo.

Finally, I figured out why the refactor was crashing when Jenkings tried to build the OoT repo. The `cfg` pointer was not being properly initialized if no config file was provided, so it would have a null pointer instead. ZAPD willl try to access the actor/object list by dereferencing that null pointer and, for surprise of nobody, crash. I changed a few things, so if no config file is provided then ZAPD will just use the hex value for the actors/objects instead.

PR #109, opened 2021-04-04
Add MM scene support

Rozelette:
This adds handling for the new scene commands added in MM. A notable exception is that MM cutscenes are not parsed yet due to RE still needing to be done. Also, the names for the commands and structs are what MM currently uses, but are subject to change, so there may be addition PRs with name churn.

Also I ran clang-format but it was a bit overzealous, so I apologize for the whitespace changes.


ZAPD/ZRoom/Commands/SetActorCutsceneList.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+		currentPtr += 16;
+	}
+	
+	string declaration = "";
+
+	for (ActorCutsceneEntry* entry : cutscenes)
+	{
+		declaration +=
+			StringHelper::Sprintf("\t{ %i, %i, %i, %i, %i, %i, %i, %i, %i, %i },\n", entry->priority,
```
AngheloAlf: We try to use 4 spaces instead of a tab for the output.
Please change, unless the MM codebase use tabs instead of spaces too.
```suggestion
			StringHelper::Sprintf("    { %i, %i, %i, %i, %i, %i, %i, %i, %i, %i },\n", entry->priority,
```

ZAPD/ZRoom/Commands/SetActorList.cpp
```diff
@@ -104,7 +98,7 @@ string SetActorList::GenerateSourceCodePass2(string roomName, int baseAddress)
 	}
 
 	zRoom->parent->AddDeclarationArray(
-		segmentOffset, DeclarationAlignment::None, DeclarationPadding::Pad16, actors.size() * 16,
+		segmentOffset, DeclarationAlignment::None, DeclarationPadding::None, actors.size() * 16,
```
AngheloAlf: Is this true for OoT and MM?

ZAPD/ZRoom/Commands/SetMesh.cpp
```diff
@@ -355,7 +355,7 @@ void SetMesh::GenDListDeclarations(std::vector<uint8_t> rawData, ZDisplayList* d
 	for (pair<uint32_t, string> vtxEntry : dList->vtxDeclarations)
 	{
 		zRoom->parent->AddDeclarationArray(
-			vtxEntry.first, DeclarationAlignment::Align8,
+			vtxEntry.first, DeclarationAlignment::None,
```
AngheloAlf: Why?

ZAPD/ZRoom/Commands/SetMinimapChests.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "../../ZFile.h"
+#include "../ZRoom.h"
+
+using namespace std;
+
+SetMinimapChests::SetMinimapChests(ZRoom* nZRoom, std::vector<uint8_t> rawData, int rawDataIndex)
+	: ZRoomCommand(nZRoom, rawData, rawDataIndex)
+{
+	int numChests = rawData[rawDataIndex + 1];
+	segmentOffset = BitConverter::ToInt32BE(rawData, rawDataIndex + 4) & 0x00FFFFFF;
```
AngheloAlf: Try to use the `GETSEGOFFSET` macro or the `Seg2Filespace` function in situations like this.
```suggestion
	segmentOffset = GETSEGOFFSET(BitConverter::ToInt32BE(rawData, rawDataIndex + 4));
```

ZAPD/ZRoom/Commands/SetMinimapList.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "../../ZFile.h"
+#include "../ZRoom.h"
+
+using namespace std;
+
+SetMinimapList::SetMinimapList(ZRoom* nZRoom, std::vector<uint8_t> rawData, int rawDataIndex)
+	: ZRoomCommand(nZRoom, rawData, rawDataIndex)
+{
+	segmentOffset = BitConverter::ToInt32BE(rawData, rawDataIndex + 4) & 0x00FFFFFF;
+	listSegmentOffset = BitConverter::ToInt32BE(rawData, segmentOffset + 0) & 0x00FFFFFF;
```
AngheloAlf: Same here with `GETSEGOFFSET` or `Seg2Filespace`.

ZAPD/ZFile.cpp
```diff
@@ -852,9 +852,19 @@ string ZFile::ProcessDeclarations()
...
+			{
+				item.second->size++;
+			}
+		}
+		else
+		{	
+			while (item.second->size % 4 != 0)
+			{
+				item.second->size++;
+			}
```
AngheloAlf: Does the current `Alignment` always align the data to `4`, even if the `DeclarationAlignment` is set to `None`?
(This question isn't specific to this PR).

ZAPD/ZCutsceneMM.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+	string output = "";
+	size_t size = 0;
+	int32_t curPtr = 0;
+
+	output += StringHelper::Sprintf("\tCS_BEGIN_CUTSCENE(%i, %i),\n", numCommands, endFrame);
+
+	for (size_t i = 0; i < data.size(); i++)
+	{
+		output += StringHelper::Sprintf("\t0x%08X,", data[i]);
```
AngheloAlf: As said earlier, spaces instead of tabs :D
```suggestion
	output += StringHelper::Sprintf("    CS_BEGIN_CUTSCENE(%i, %i),\n", numCommands, endFrame);

	for (size_t i = 0; i < data.size(); i++)
	{
		output += StringHelper::Sprintf("    0x%08X,", data[i]);
```

AngheloAlf (changes requested):
Nice job, just a few minor comments.
Btw, Jenkins is having troubles checking this PR.

ZAPD/ZRoom/Commands/SetActorCutsceneList.cpp
fig02: hopefully in the future we will be able to use OutputFormatter to do this automatically. that way we can use \t in the code which looks a bit cleaner and then change it to spaces later

fig02:
theres def something wrong with clang here lol, the whitespace changes are wild
we have this same issue in oot where the results of clang format can be different for different people. its really strange. hopefully we can figure out the issue

ZAPD/ZRoom/Commands/SetActorList.cpp
Rozelette: I assume OOT actor lists were the end of a separate .c file, so they were padded out to 16-bytes. In MM, it seems some other structs could be at the end of them, so this padding isn't always correct.

ZAPD/ZRoom/Commands/SetMesh.cpp
Rozelette: I'm not sure, by I encountered several instances in MM where these were not aligned to 8-bytes

ZAPD/ZFile.cpp
Rozelette: I think it does, which is why I had to add the `Align2` value. I would assume that `None` would not cause any alignment, but then I encountered trouble using it for `Vec3s[]`s. This does seem unintuitive to me, but I didn't want to change that behavior.

ZAPD/ZRoom/Commands/SetActorList.cpp
AngheloAlf: I'm not sure if the padding is really needed in OoT, but in the meanwhile you could add a game check.
```c
        DeclarationPadding padding = DeclarationPadding::Pad16;
        if (Globals::Instance->game == MM_RETAIL)
        {
            padding = DeclarationPadding::None;
        }
        zRoom->parent->AddDeclarationArray(
		segmentOffset, DeclarationAlignment::None, padding, actors.size() * 16,
```

ZAPD/ZRoom/Commands/SetMesh.cpp
AngheloAlf: You could add the same check here too.

ZAPD/ZFile.cpp
AngheloAlf: Well, this should be fixed in a future PR. I'll add a new issue in the meanwhile.

AngheloAlf (approved):

NEstelami (approved):

PR #116, opened 2021-04-16
Deprecate Hint system

AngheloAlf:
- Remove `BlobHint`, `TextureHint` and `BackgroundHint`.
- Add a "deprecation warning". It will trigger if any of the "Hint" nodes are found in a xml file.
- Fix the `Cutscene` node, so it can be used instead of `CutsceneHint`.

The idea behind this PR is to allow an easier migration for the already existing xmls in the repo.


NEstelami:
Looks good. I think people are going to need to update their XMLs for this one since most of the hints are outright removed, right?

AngheloAlf:
Is safe to remove the hints I removed, because
- `BlobHint` does exactly the same as `unaccounted`, it doesn't even extract it as a `.bin` file.
- `TextureHint` is not being used anywhere in the repo and it doesn't work properly either.
- `BackgroundHint` is not even in the OoT repo right now.

About the other `Hint`s, I'm planning to do a PR in OoT to change them to their non-`Hint` version, so they could be safely removed from ZAPD in the future.

PR #121, opened 2021-04-22
Moved BG Screen Width and Height into Config File

NEstelami:
The following can be added to Config.xml to change these values:

```xml
<BGConfig ScreenWidth="320" ScreenHeight="240"/>
```

If the `BGConfig` node is not present, or if `ScreenWidth`/`ScreenHeight` are not defined, it will default to 320x240.


AngheloAlf (commented):
Looks good.
I suggest adding the `BgConfig` node to the config files when they get added to their respectives repos, so people would know this exists.

NEstelami:
Sure thing. I'll add it to my `oot` PR.

Issue #18, opened 2020-12-29
ZAPD needs more robust error handling


Kenix3:
In many cases when ZAPD loads a file and the file can't be found, it will just crash. It should output a message like "file not found."
- [ ] ZAPD should also support XML file validation per asset type.
- [x] ZAPD should error if an invalid ZResource name is included in the XML.
- [ ] ZAPD should error when a ZResource is not in a ZFile.
- [ ] ZAPD should error when a ZFile is encountered inside of a ZFile.
- [ ] ZAPD should error when a ZResource node has a child.
- [ ] ZAPD should error when invalid ZResource attributes and attribute values are provided.
- [x] ZAPD should error if parameters are out of range. For example, a size parameter larger than the type used to store it in ZAPD.

AngheloAlf:
I have a few questions about this:
- "ZAPD should also support XML file validation per asset type." What does this mean? Does it means ZAPD should check if an the attributes of an asset is valid? 
- "ZAPD should error if an invalid ZResource name is included in the XML." I think ZAPD is already checking this [here](https://github.com/zeldaret/ZAPD/blob/c4773301a17a5a9f9c964a3e7ba6dfbf0730a94b/ZAPD/ZFile.cpp#L409).

PR #126, opened 2021-04-28
Check XML resource attributes

AngheloAlf:
ZAPD will now perform more runtime checks for the attributes of each declared resource in the XMLs.
- Checks if an invalid/unknown attribute was used in a resource.
  - Examples:
  ```
  Exception occurred: ZResource::ParseXML: Fatal error while parsing 'object_spot09_obj'.
	 Unexpected 'Size' attribute in resource 'Collision'.
	 Aborting...
  ```
  ```
  Exception occurred: ZResource::ParseXML: Fatal error while parsing 'object_sd'.
	 Unexpected 'Nane' attribute in resource 'Limb'.
	 Aborting...
  ```
- Checks if a resource is missing a required/mandatory attribute.
  - Example:
  ```
  Exception occurred: ZResource::ParseXML: Fatal error while parsing 'object_gi_saw'.
         Missing required attribute 'Name' in resource 'DList'.
         Aborting...
  ```
- General cleanup.
- Change the internal way of `ZArray` to extract multiples elements. It doesn't need to create a new `ZFile` anymore.
- To perform the checks, each `ZResource` needs to register the valid attributes for each one of them.
  - To do this, 2 new methods are provided.
    - `RegisterRequiredAttribute(attr)`: Defines a mandatory attribute. ZAPD will throw an exception if a required attribute is missing. It's value will be stored in `requiredAttributes`.
    - `RegisterOptionalAttribute(attr, defaultValue = "")`. Defines an optional attribute. Each `ZResource` must check if the attribute is declared or not. It's value will be stored in `optionalAttributes`.
  - The valid attributes should be registered in the constructor of each `ZResource`
  - Every `ZResource` will register the required the following attributes automatically:
    - `Name` as required.
      - If it is an inner node (for example, subnode of `ZArray`), then 'Name' isn't required.
    - `OutName` and `Offset` as optional.
    - `Custom` as non-value.
- Check if `Width` and `Height` attributes of `ZTexture` has only decimal digits, and throw an exception otherwise.

Related PR: #18 

This PR will need a few changes in the OoT repo, since some objects are "invalid". ~~I'll PR the needed changes soon~~ [PR 799](https://github.com/zeldaret/oot/pull/799).


fig02 (commented):
Everything is good oot-side for this now.
This would be good to get in when possible so we can move forward with https://github.com/zeldaret/oot/pull/811

fig02 (approved):

NEstelami (approved):

PR #127, opened 2021-04-30
Exporter Support Added

NEstelami:
In the interest of making things more modular and future proofing, especially for mod tools, a "exporter" system has been implemented. 
An exporter "hooks" into a certain ZResource type and can write data of its own in whatever format it is designed for.
Exporters can be created inside ZAPD itself or as a static library. If created as a static library, it will automatically be "registered" when linked with ZAPD without modification of ZAPD's code.
It's possible to have more than one set of exporters registered and to use a command line option to set which one gets used.

An exporter can also add in new file modes and command line arguments.

I have included an example exporter to give an idea of how this system works. It just takes certain ZResources and serializes them back to their binary format.

The "HL" code has also been removed. It's going to be converted into an exporter at a later time.


ZAPD/ZRoom/Commands/SetEntranceList.h
```diff
@@ -29,5 +30,6 @@ class SetEntranceList : public ZRoomCommand
 	std::string GetCommandCName() const override;
 
 private:
-	std::vector<EntranceEntry> entrances;
+	std::vector<uint8_t> _rawData;
+	int32_t _rawDataIndex;
```
AngheloAlf: This looks unused.

ZAPD/ZRoom/Commands/SetCsCamera.h
```diff
@@ -37,6 +40,6 @@ class SetCsCamera : public ZRoomCommand
 	std::string GetCommandCName() const override;
 
 private:
-	std::vector<CsCameraEntry> cameras;
-	std::vector<ZVector> points;
+	std::vector<uint8_t> _rawData;
+	int32_t _rawDataIndex;
```
AngheloAlf: Unused

ZAPD/ZResource.h
```diff
@@ -7,8 +7,17 @@
...
-#include "Directory.h"
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace fs = std::filesystem;
+#else
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#endif
+
+//#include "Directory.h"
```
AngheloAlf: Why have this directly instead of simply `#include "Directory.h"`?

ZAPD/ZFile.h
```diff
@@ -3,21 +3,21 @@
...
 #include "tinyxml2.h"
 
 enum class ZFileMode
 {
 	BuildTexture,
 	BuildOverlay,
-	BuildModelIntermediette,
-	BuildAnimationIntermediette,
+	//BuildModelIntermediette,
+	//BuildAnimationIntermediette,
```
AngheloAlf: Remove

ZAPD/ZRoom/ZRoomCommand.h
```diff
@@ -79,8 +81,6 @@ class ZRoomCommand : public ZResource
 protected:
 	ZRoom* zRoom;
 
-	RoomCommand cmdID;
 	uint8_t cmdArg1;
 	segptr_t cmdArg2;
```
AngheloAlf, line 87: I guess you'll like to make those public too.

ZAPD/packages.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="libpng" version="1.6.28.1" targetFramework="native" />
+  <package id="libpng.redist" version="1.6.28.1" targetFramework="native" />
+  <package id="zlib" version="1.2.8.8" targetFramework="native" />
+  <package id="zlib.v120.windesktop.msvcstl.dyn.rt-dyn" version="1.2.8.8" targetFramework="native" />
+  <package id="zlib.v140.windesktop.msvcstl.dyn.rt-dyn" version="1.2.8.8" targetFramework="native" />
+</packages>
```
AngheloAlf, line 8: What is this file for? VisualStudio?

ZAPD/MemoryStream.cpp
```diff
@@ -0,0 +1,90 @@
...
+		bufferSize += length;
+	}
+
+	memcpy_s(&buffer[baseAddress], length, srcBuffer, length);
+	baseAddress += length;
+}
+
+void MemoryStream::WriteByte(int8_t value)
+{
+	if (baseAddress >= buffer.size())
```
AngheloAlf: This gives a warning
```
warning: comparison of integer expressions of different signedness
```

AngheloAlf (commented):
A few minor comments for now

ZAPD/packages.config
NEstelami: Yep. Nuget packages.

ZAPD/ZResource.h
NEstelami: I was experiencing some weird compilation issues before, but they look to have been resolved.

ZAPD/ZResource.h
```diff
@@ -7,6 +7,15 @@
...
 #include "tinyxml2.h"
+#include <BinaryWriter.h>
+
+//#if __has_include(<filesystem>)
+//#include <filesystem>
+//namespace fs = std::filesystem;
+//#else
+//#include <experimental/filesystem>
+//namespace fs = std::experimental::filesystem;
+//#endif
```
AngheloAlf: Remove

ZAPD/ZResource.h
```diff
@@ -124,3 +141,14 @@ typedef ZResource*(ZResourceFactoryFunc)(ZFile* nParent);
...
+#define REGISTER_EXPORTER(expFunc)																		\
+	class ZResExp_##expFunc																				\
+	{																									\
+	public:																								\
+		ZResExp_##expFunc()																			\
+		{																								\
+			##expFunc();																				\
+		}																								\
+	};																									\
+	static ZResExp_##expFunc inst_ZResExp_##expFunc;
```
AngheloAlf: This gives me the following error:
```
In file included from ./TextureExporter.h:3,
                 from Main.cpp:1:
../ZAPD/ZResource.h:150:3: error: pasting "{" and "ImportExporters" does not give a valid preprocessing token
  150 |   {                        \
      |   ^
Main.cpp:79:1: note: in expansion of macro ‘REGISTER_EXPORTER’
   79 | REGISTER_EXPORTER(ImportExporters)
      | ^~~~~~~~~~~~~~~~~

```

ExporterTest/RoomExporter.cpp
```diff
@@ -0,0 +1,373 @@
...
+			writer->Write((uint8_t)0); // 0x02
+			writer->Write((uint8_t)0); // 0x03
+			writer->Write((uint8_t)0); // 0x04
+			writer->Write((uint8_t)0); // 0x05
+			writer->Write((uint8_t)0); // 0x06
+			writer->Write((uint8_t)0); // 0x07
+		}
+		break;
+		default:
+			printf("UNIMPLEMENTED COMMAND: %i\n", cmd->cmdID);
```
AngheloAlf:
```
RoomExporter.cpp:357:36: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘RoomCommand’ [-Wformat=]
  357 |    printf("UNIMPLEMENTED COMMAND: %i\n", cmd->cmdID);
      |                                   ~^     ~~~~~~~~~~
      |                                    |          |
      |                                    int        RoomCommand
```

ZAPD/MemoryStream.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include "Stream.h"
+#include <memory>
+#include <vector>
+
+class MemoryStream : public Stream
```
AngheloAlf: What's the point of this class?

ZAPD/BinaryWriter.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "Stream.h"
+#include <memory>
+#include <array>
+#include <vector>
+#include <string>
+
+class BinaryWriter
```
AngheloAlf: Same here, why are this classes needed?
Maybe we could move classes like this one and `MemoryStream`, and some others already in the repo like `OutputFormatter`, `ImageBackend`, `StringHelper`, Directory and so on to a folder `ZAPD/Utils/` or something. What do you think?

ZAPD/MemoryStream.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+	if (seekType == SeekOffsetType::Start)
+		baseAddress = offset;
+	else if (seekType == SeekOffsetType::Current)
+		baseAddress += offset;
+	else if (seekType == SeekOffsetType::End)
+		baseAddress = bufferSize - 1 - offset;
+}
+
+std::unique_ptr<char[]> MemoryStream::Read(size_t length)
```
AngheloAlf: Why not simply return a `std::vector<char>`? Would do the same without having to think about pointers.

ZAPD/Main.cpp
```diff
@@ -22,6 +20,7 @@
 
 #include <string>
 #include "tinyxml2.h"
+//#include <HighLevel/HLExporterTest.h>
```
AngheloAlf: Remove

ZAPD/Main.cpp
```diff
@@ -53,7 +50,7 @@ void ErrorHandler(int sig)
 	fprintf(stderr, "Traceback:\n");
 	for (size_t i = 1; i < size; i++)
 	{
-		Dl_info info;
+		Dl_info info;
```
AngheloAlf: Here is a weird extra space. Don't forget to run the formatter.
```suggestion
		Dl_info info;
```

ZAPD/Main.cpp
```diff
@@ -218,18 +175,94 @@ int main(int argc, char* argv[])
 		{
 			Globals::Instance->warnUnaccounted = true;
 		}
+		else if (arg == "-sexp" || arg == "--set-exporter")  // Set Current Exporter
```
AngheloAlf: I'm not a fan of `sexp`. What about `-se` instead?

ZAPD/Globals.cpp
```diff
@@ -176,11 +177,43 @@ void Globals::AddSegment(int32_t segment, ZFile* file)
...
+
+void Globals::AddExporter(std::string exporterName, ExporterSet* exporterSet)
+{
+	std::map<std::string, ExporterSet*>* exporters = GetExporterMap();
+	(*exporters)[exporterName] = exporterSet;
+}
+
+ZResourceExporter* Globals::GetExporter(ZResourceType resType)
+{
+	auto exporters = *GetExporterMap();
```
AngheloAlf: Doing this would make a copy of the whole Exporter Map, which would be a waste of resources imo.
What about using the pointer instead? 
```suggestion
	auto exporters = GetExporterMap();
```
If you don't want to let the user modify the original, you could either return a `const` pointer or a `const` reference.

ZAPD/Globals.cpp
```diff
@@ -176,11 +177,43 @@ void Globals::AddSegment(int32_t segment, ZFile* file)
...
+
+	if (currentExporter != "" && exporters[currentExporter]->exporters.find(resType) != exporters[currentExporter]->exporters.end())
+		return exporters[currentExporter]->exporters[resType];
+	else
+		return nullptr;
+}
+
+ExporterSet* Globals::GetExporterSet()
+{
+	auto exporters = *GetExporterMap();
```
AngheloAlf: Same here.

AngheloAlf (changes requested):
I couldn't compile this because it threw an error.
I compiled with the following command:
```
g++ *.cpp -I ../ZAPD/ -I ../lib/tinyxml2/ --std=c++17 -I .
```

Also, could you add a `Makefile` for this?

ZAPD/Main.cpp
NEstelami: `-se` could work. Hopefully that won't collide with any other args in the future.

ZAPD/BinaryWriter.h
NEstelami: These classes help with reading/writing binary data in a C#-like way. I agree it should probably be put in a utils folder or even made its own library (since these classes would be useful beyond just ZAPD itself).

ZAPD/ZResource.h
NEstelami: Seems there were some differences between MSVC and G++. The latest push should fix it.

ZAPD/ZAnimation.cpp
```diff
@@ -26,14 +25,6 @@ void ZAnimation::ParseRawData()
...
 {
-	if (Globals::Instance->testMode)
-	{
-		HLAnimationIntermediette* anim = HLAnimationIntermediette::FromZAnimation(this);
-		std::string xml = anim->OutputXML();
-		File::WriteAllText(outFolder / (name + ".anmi"), xml);
-
-		delete anim;
-	}
 }
```
AngheloAlf: Wouldn't be a better idea to simply remove the whole function instead of leave it empty?

ZAPD/ImageBackend.cpp
```diff
@@ -216,7 +216,7 @@ void ImageBackend::WritePng(const char* filename)
 
 void ImageBackend::WritePng(const fs::path& filename)
 {
-	WritePng(filename.c_str());
+	WritePng((char*)filename.c_str());
```
AngheloAlf: Why is this cast needed? `.c_str()` should be returning a `const char*`, which is the expected type for that method, so I'm curious about this.

Makefile
```diff
@@ -74,5 +74,8 @@ ZAPD/Main.o: genbuildinfo ZAPD/Main.cpp
 lib/libgfxd/libgfxd.a:
 	$(MAKE) -C lib/libgfxd
 
-ZAPD.out: $(O_FILES) lib/libgfxd/libgfxd.a
-	$(CXX) $(CXXFLAGS) $(INC) $(O_FILES) lib/libgfxd/libgfxd.a -o $@ $(FS_INC) $(LDFLAGS)
+ExporterTest/ExporterTest.a:
+	$(MAKE) -C ExporterTest
+
+ZAPD.out: $(O_FILES) lib/libgfxd/libgfxd.a ExporterTest/ExporterTest.a
+	$(CXX) $(CXXFLAGS) $(INC) $(O_FILES) lib/libgfxd/libgfxd.a -Wl,--whole-archive ExporterTest/ExporterTest.a -Wl,--no-whole-archive -o $@ $(FS_INC) $(LDFLAGS)
```
AngheloAlf, line 81: I'm no expert, but using `-Wl,--whole-archive` and `-Wl,--no-whole-archive` feels a bit contradictory. Why it is needed for?

ExporterTest/Makefile
```diff
@@ -0,0 +1,23 @@
+CFLAGS = -Wall -O2 -g -std=c++17
```
AngheloAlf: Since this is a C++ thing, this should be CXXFLAGS, right?
```suggestion
CXXFLAGS = -Wall -O2 -g -std=c++17
```

ZAPD/Globals.cpp
```diff
@@ -177,11 +178,44 @@ void Globals::AddSegment(int32_t segment, ZFile* file)
...
+
+std::map<std::string, ExporterSet*>& Globals::GetExporterMap()
+{
+	static std::map<std::string, ExporterSet*> exporters;
+	return exporters;
+}
+
+void Globals::AddExporter(std::string exporterName, ExporterSet* exporterSet)
+{
+	std::map<std::string, ExporterSet*>& exporters = GetExporterMap();
```
AngheloAlf: you can use `auto` here if you like.
```suggestion
	auto exporters = GetExporterMap();
```

ZAPD/Main.cpp
```diff
@@ -253,18 +209,93 @@ int main(int argc, char* argv[])
 		{
 			Globals::Instance->verboseUnaccounted = true;
 		}
+		else if (arg == "-se" || arg == "--set-exporter")  // Set Current Exporter
```
AngheloAlf, line 212: Could you document this flag in the README.md?

Makefile
NEstelami: --whole-archive is needed otherwise `REGISTER_EXPORTER` gets optimized out. --no-whole-archive is needed to prevent everything that comes after `ExporterTest.a` from being affected by `--whole-archive`

AngheloAlf: Ohh, so order matters here, I see.

ZAPD/ImageBackend.cpp
NEstelami: To be honest I don't remember why that change was made. I would think it was a compiler issue but removing it works fine so your guess is as good as mine.

AngheloAlf (approved):

PR #128, opened 2021-05-02
Background array declarations now use string literal

NEstelami:


ZAPD/ZFile.cpp
```diff
@@ -935,13 +899,22 @@ string ZFile::ProcessDeclarations()
...
+						item.second->varName.c_str(),
+						item.second->arrayItemCntStr.c_str());
+				}
 				else
-					output += StringHelper::Sprintf("%s %s[%i];\n", item.second->varType.c_str(),
-					                                item.second->varName.c_str(),
-					                                item.second->arrayItemCnt);
+				{
+					if (item.second->arrayItemCnt == 0)
+						output += StringHelper::Sprintf("%s %s[];\n", item.second->varType.c_str(),
```
AngheloAlf: What about merging this as only one `else if`?

AngheloAlf (commented):
A minor comment. Besides that, looks good.

AngheloAlf:
Don't forget to fix the conflict too.

Issue #113, opened 2021-04-13
Static can't be manually added to an asset


louist103:
Static should have a manual enable for things like overlay assets.

AngheloAlf:
This should be done too

PR #129, opened 2021-05-02
(DRAFT) Added support for manually setting an asset XML node as static.

NEstelami:
There are still some bugs with this I need to sort out, along with code cleanup. The reason why I'm putting this up in the first place is just to make clear that this is in the works so that work is not duplicated.

### DIFF
--- a/ZAPD/ZDisplayList.cpp
+++ b/ZAPD/ZDisplayList.cpp
@@ -1953,6 +1953,8 @@ string ZDisplayList::GetSourceOutputCode(const std::string& prefix)
 		{
 			string declaration = "";
 
+			item.second->isStatic = isStatic;
+
 			declaration += item.second->GetSourceOutputCode(prefix);
 			texDeclarations[item.first] = declaration;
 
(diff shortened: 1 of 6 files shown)


AngheloAlf:
Related issue: #113

ZAPD/ZDisplayList.cpp
```diff
@@ -2294,7 +2294,7 @@ std::string ZDisplayList::GetExternalExtension()
 
 std::string ZDisplayList::GetSourceTypeName()
 {
-	return "Gfx";
+	return isStatic ? "static Gfx" : "Gfx";
```
AngheloAlf, line 2297: This part seems a bit unnecessary , because ZFile seems to handle the `declType` by itself, right?

ZAPD/ZTexture.cpp
```diff
@@ -876,7 +876,7 @@ ZResourceType ZTexture::GetResourceType()
 
 std::string ZTexture::GetSourceTypeName()
 {
-	return "u64";
+	return isStatic ? "static u64" : "u64";
```
AngheloAlf, line 879: Same here.

ZAPD/ZVtx.cpp
```diff
@@ -25,7 +25,7 @@ void ZVtx::ParseXML(tinyxml2::XMLElement* reader)
 
 std::string ZVtx::GetSourceTypeName()
 {
-	return "Vtx";
+	return isStatic ? "static Vtx" : "Vtx";
```
AngheloAlf, line 28: Same here.

ZAPD/ZDisplayList.cpp
NEstelami: Good point. ZFile might be able to handle this itself since it knows what ZResources are flagged as static.

Issue #13, opened 2020-12-29
ZAPD should support more object file types


Kenix3:
Kenix
- [x] string
- [x] Vector (vec3s, vec3f, vec3i)
- [x] Primitive (u8/s8, u16/s16, u32/s32, u64/s64, f32, f64)

Jack

- [x] arrays
- [x] Standard skeletons
- [x] Flex skeletons
- [x] Standard limbs
- [x] LOD limbs
- [x] Skin limbs
- [x] Link animation headers
- [x] Legacy animation system (object_human)

NEstelami:
Is `string` still something we want? Do we know of cases where it would be useful?

AngheloAlf:
ZAPD already supports everything listed here. So I think this issue can be closed now.

PR #130, opened 2021-05-02
Add `ZString` resource

AngheloAlf:
Adds a `<String>` XML node for string extraction.

Related PR: #13


ZAPD/ZString.cpp
```diff
@@ -0,0 +1,58 @@
...
+ZString::ZString(ZFile* nParent) : ZResource(nParent)
+{
+}
+
+void ZString::ParseRawData()
+{
+	size_t size = 0;
+    for (size_t i = rawDataIndex; i < rawData.size(); ++i) {
+        ++size;
+        if (rawData.at(i) == '\0') {
```
NEstelami: I would recommend getting a `uint8_t*` pointer from `rawData.data()` and iterating through this loop using that pointer. Calling `at()` or using an array operator on `rawData` tends to have a performance impact if done frequently.

ZAPD/ZString.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+	parent->AddDeclarationArray(rawDataIndex, DeclarationAlignment::None, GetRawDataSize(),
+	                            GetSourceTypeName(), name, 0, GetBodySourceCode());
+
+	return "";
+}
+
+std::string ZString::GetSourceOutputHeader(const std::string& prefix)
+{
+	return StringHelper::Sprintf("#define %s_macro \"%s\"", name.c_str(), rawData.data());
```
NEstelami, line 44: Could you explain what this macro is for?

AngheloAlf: I added this macro as a way to extract the string without actually defining it as a variable.
It could be useful if somebody wants to extract the strings used in overlay's rodata.

NEstelami: Gotcha.

ZAPD/ZString.cpp
AngheloAlf: I think using `rawData[]` should have the same performance impact as using a `uint8_t*`, but I don't have proof, so I'll do it like you said.

NEstelami: When I was doing profiling looking for hotspots in ZAP's code, it was one thing I spotted. I have not looked at the std::vector implementation (and it may vary from compiler to compiler), but I think it might be doing extra work.

PR #131, opened 2021-05-03
Fix `-export-dynamic` in mac

AngheloAlf:
`-Wl,-export-dynamic` doesn't work well in Mac/clang.
This was needed to unmangle the traceback, since it isn't a critical feature, then it will be disabled for them.


ethteck (approved):
Tested, and this fixes the build in mac! thanks :)

PR #133, opened 2021-05-05
Change image library to `libpng` and improve how palettized textures are extracted

AngheloAlf:
- Change the library used for PNG files. From `stb` to `libpng`.
- Include the palette inside palettized textures (`ci4` and `ci8`).
  - Currently this palette will be ignored when building back the PNG to a C array until we figure out a better solution.
- Add `TlutOffset` attribute to `<Texture>` node.
  - This should be used only in `ci4` or `ci8` textures.
  - Is the offset of the used TLUT by the current texture.
  - ZAPD will try to automatically assign each TLUT to their corresponding `ci` textures based on their use in the extracted DLists.
- Improve TLUT's overlapping handling.
- Now `ZFile` is in charge of tracking textures, as oposed of `ZDisplayList`.
- Adds a `README.md`.
- Makefile: use `make ASAN=1` to compile with Address sanitizer enabled.
- Add easter egg.
- Adds `-crc`/`-output-crc` parameter. If it is provided when invoking ZAPD, then ZAPD will create a file per texture containing its CRC hash.
- Remove redundant `CalcHash` calls.
- Probably something else I forgot :sweat_smile:


NEstelami:
Is `CalcHash()` no longer called on all resources?

AngheloAlf:
No. Are we interested in doing that?
I removed it since it was only used for textures, but I could re-add it.

NEstelami:
Right now we're not calculating hashes for non-textures but it may be useful in the future. The function is virtual so it doesn't do any actual CRC calculations for non-textures and won't impact performance.

AngheloAlf:
I changed it back to always calculate the hash for every extracted asset.

NEstelami (approved):

PR #136, opened 2021-05-06
Refactor: Removed `relPath` since it is no longer used.

NEstelami:


AngheloAlf (commented):
You should remove `std::string relativePath;` from `ZResource.h` too, right?

NEstelami:
> You should remove `std::string relativePath;` from `ZResource.h` too, right?

Thought I did. Must have missed it.

AngheloAlf (approved):

Issue #135, opened 2021-05-06
Add MM to Jenkins


AngheloAlf:
Recently MM added ZAPD to their compilation process, so we should be checking we don't break MM support.

PR #124, opened 2021-04-24
ZRoom refactor

AngheloAlf:
- Clean up a lot of ZRoom/Commands code.
- Export the rooms/scenes commands as an array of `SCmdBase`. The output for those arrays use the macros made by @Zelllll 
- General cleanup of rooms/scenes outputted code.
- ~~Automatically adjust TLUTs to be smaller in case of intersections.~~
- Remove a lot of unnecessary pointers and destructors.
- Properly implement `ZPath`.
- Move the class `Declaration` to its own file.
- Probably something else I don't remember right now.

Because of macros being needed for rooms commands, this will not build properly by Jenkins. I'll do a PR to add the new macros to the OoT repo soon :tm:

Examples:

Current command output:
```c
static SCmdSoundSettings bowling_sceneSet0000Cmd00 = {  0x15, 0x05, 0x00, 0x00, 0x00, 0x00, 0x13, 0x4E }; // 0x0000

static SCmdRoomList bowling_sceneSet0000Cmd01 = {  0x04, 0x01, (u32)&bowling_sceneRoomList0x000080 }; // 0x0008

static SCmdTransiActorList bowling_sceneSet0000Cmd02 = {  0x0E, 0x01, (u32)bowling_sceneTransitionActorList0x000070 }; // 0x0010

static SCmdMiscSettings bowling_sceneSet0000Cmd03 = {  0x19, 0x00, 0x0000000E }; // 0x0018

static SCmdColHeader bowling_sceneSet0000Cmd04 = {  0x03, 0x00, (u32)&bowling_sceneCollisionHeader0x001A74 }; // 0x0020

static SCmdEntranceList bowling_sceneSet0000Cmd05 = {  0x06, 0x00, (u32)&bowling_sceneEntranceList0x000088 }; // 0x0028

static SCmdSpecialFiles bowling_sceneSet0000Cmd06 = {  0x07, 0x00, 0x0002 }; // 0x0030

static SCmdSpawnList bowling_sceneSet0000Cmd07 = {  0x00, 0x01, (u32)&bowling_sceneStartPositionList0x000060 }; // 0x0038

static SCmdSkyboxSettings bowling_sceneSet0000Cmd08 = {  0x11, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 }; // 0x0040

static SCmdExitList bowling_sceneSet0000Cmd09 = {  0x13, 0x00, (u32)&bowling_sceneExitList0x00008C }; // 0x0048

[...]

PR #137, opened 2021-05-07
Add MM build target to Jenkinsfile

AngheloAlf:
Jenkins now should checkout the latest MM and test it, so we don't broke compatibility with them.

Issue: #135

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,7 +49,7 @@ pipeline {
         stage('Set up mm') {
             steps {
                 dir('mm') {
-                    sh 'cp /usr/local/etc/roms/mm.us.rev1.z64 baserom.z64'
+                    sh 'cp /usr/local/etc/roms/mm.us.rev1.z64 baserom.mm.us.rev1.z64'
 
                     // Identical to `make setup` except for copying our newer ZAPD.out into mm
                     sh 'git submodule update --init --recursive'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,9 +51,8 @@ pipeline {
                 dir('mm') {
                     sh 'cp /usr/local/etc/roms/mm.us.rev1.z64 baserom.z64'
 
-                    // Identical to `make setup` except for copying our newer ZAPD.out into oot
+                    // Identical to `make setup` except for copying our newer ZAPD.out into mm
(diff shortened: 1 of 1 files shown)


AngheloAlf:
Looks like MM failed to build because of some issues with the `OutputFormatter`

There is output like this in MM scenes with current ZAPD's master:
```c
static SCmdCsCameraList Z2_SYATEKI_MORISet0000Cmd0D = { 0x02, 1, (u32)&Z2_SYATEKI_MORICsCameraList0x000574 }; // 0x0068
    }; 

static SCmdMinimapSettings Z2_SYATEKI_MORISet0000Cmd0E = { 0x1C, 0, (u32)&Z2_SYATEKI_MORIMinimapList0x000558 }; //
    0x0070 }; 
```

Problems like this are fixed in #124. but that branch wont build properly either because of the lack of the new macros for scenes commands (https://github.com/zeldaret/mm/pull/114).

So I'll convert this PR into a draft until those two PRs get merged in their respective repos.

Jenkinsfile
```diff
@@ -38,6 +38,35 @@ pipeline {
...
+            }
+        }
+        stage('Set up mm') {
+            steps {
+                dir('mm') {
+                    sh 'cp /usr/local/etc/roms/mm.us.rev1.z64 baserom.z64'
+
+                    // Identical to `make setup` except for copying our newer ZAPD.out into oot
+                    sh 'git submodule update --init --recursive'
+                    sh 'python3 -m pip install -r requirements.txt'
```
ethteck: We shouldn't be installing any dependencies at build time. What's missing that requires this?

AngheloAlf: I copy-pasted that from MM's makefile. From the setup step to be precise.
https://github.com/zeldaret/mm/blob/aae6d9d3644f0792a27b3c67758345c6d13e909b/Makefile#L196

I guess I should remove this, right?

ethteck: Ah okay. My bad for missing it there then. Yeah I think it can be removed

AngheloAlf:
MM updated their ZAPD version, so now this properly works.
Would be nice to prioritize this PR to prevent breaking MM compatibility in the future.

fig02 (approved):

NEstelami (approved):
LGTM

PR #138, opened 2021-05-09
Improve Makefile compatability (GCC 5.4, Clang 12)

zbanks:
Tested with GCC 5.4 & Clang 12 on Ubuntu 16.04

```
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ clang-12 --version
Ubuntu clang version 12.0.0-++20210319082646+4990141a4366-1~exp1~20210319193401.61
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

Tested with:
```
$ make clean
$ make -j4 rebuild
$ make clean
$ make CC=clang-12 -j4 rebuild
```

---

Overall these were some small changes I had made locally, I thought I was just a weirdo running such an old Ubuntu version, but it sounds like I'm not alone! I think these changes are pretty small, so I don't think they'll cause issues with other platforms (but I'm happy to wait to check before merging).

- This old version of GCC *requires* that `static_assert(...)` take a second message parameter, whereas it is optional in proper C++17.
- Only pass linker flags during the linking steps, not when compiling `.o`s. (I don't think it's harmful, but it produces warnings on `clang`)
- Explicitly link against `-libstdc++` and `-lm` (needed by clang, which doesn't implicitly link against them)
- Removed `-rdynamic`, which is equivalent to `-Wl,-export-dynamic` but is not supported by clang. (Can we just remove this entirely? I couldn't find why this was needed... `-g3 -ggdb` should be enough to help with backtraces?)


AngheloAlf (commented):
Looks good to me.
More testing in other archs/OSes would be nice (like aarch64 or Mac)

AngheloAlf:
Also, why do we have both `-g3` and `--ggdb`? Those can be used together? I thought one disabled the other.

zbanks:
> Also, why do we have both `-g3` and `--ggdb`? Those can be used together? I thought one disabled the other.

Yeah great question actually -- I usually use `-ggdb3` too. The [help page](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html) seems to suggest your reading of `-ggdblevel`, but I got different file sizes for `ZAPD.out` when compiling with `-g1 -ggdb` and `-g3 -ggdb`, so maybe it's equivalent? ¯\\\_(ツ)\_/¯

ethteck:
Build passes on my mac (Big Sur 11.3.1)!

ethteck (approved):

louist103:
Build passes on my Pi4.

louist103 (approved):

AngheloAlf (approved):

AngheloAlf:
This PR has a conflict!

zbanks:
Thanks for the reminder! Merged & tested on my Linux machine with gcc 5.4 & clang 12.

fig02 (approved):

AngheloAlf (approved):

NEstelami (approved):

PR #139, opened 2021-05-11
Hotfix: fix AnimatedMaterial and Minimap extraction

AngheloAlf:
The extraction of this assets broke in some PR merge. This went unnoticed because the broken assets are MM-only.

This PR is needed for [MM's PR 129](https://github.com/zeldaret/mm/pull/129)

### DIFF
--- a/copycheck.py
+++ b/copycheck.py
@@ -2,6 +2,6 @@ import os
 from shutil import copyfile
 
 if (os.environ.get('ZAPD_COPYDIR') != None):
-    print("Copying ZAPD.out to OoT repo...")
+    print("Copying ZAPD.out to repo...")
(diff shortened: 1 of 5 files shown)


copycheck.py
```diff
@@ -2,6 +2,6 @@
 from shutil import copyfile
 
 if (os.environ.get('ZAPD_COPYDIR') != None):
```
NEstelami, line 4: `ZAPD_COPYDIR` might need to be renamed to `ZAPD_OOTCOPYDIR` with an accompanying `ZAPD_MMCOPYDIR`. But that'll be for another time.

NEstelami (approved):

PR #140, opened 2021-05-11
ZRoom cleaning 3

AngheloAlf:
The third and (hopefully) the last `ZRoom` cleaning PR.

- Add ~~2~~ 1 new XML resource: ~~`SceneAltHeader` and `RoomAltHeader`~~ `AltHeader` .
  - Added to replace the old `AltHeaderHint`.
- Completely remove the `Hint` system (Finally).
- Simplify a bit the command's extraction logic.
- Add a `HackMode` attribute to `Scene`s and `Room`s to standardize a cleaner way to handle the edge cases.
- General code cleanup.
- Remove unused `json` library.
- I hope nothing else.


fig02 (approved):

fig02 (approved):

NEstelami:
Does behavior differ in any way between using `SceneAltHeader` and `RoomAltHeader`?

AngheloAlf:
From what I remember (and a quick look over the code), both do exactly the same right now. I added them as 2 nodes instead of simply one simply as a future-proof decision were we may want/have to differentiate them.
I can change them to be simply one `AltHeader` node if you prefer

NEstelami:
I think for future proofing, we can have the alt header check its parent to see if its part of a room or a scene. I'm concerned that having different ones may introduce user error if someone accidentally puts a scene alt header in a room or a room alt header in a scene and we later introduce differences based on which header you use.

AngheloAlf:
Okey, I just unified them into only one `AltHeader` node.

PR #142, opened 2021-05-13
Add `rgba16` format and deprecate `rgb5a1`

AngheloAlf:
- Adds support for `rgba16` texture format.
- Extracted PNGs of this format will have the extension `.rgba16.png`
  - This change will need support on the OoT repo (see [814](https://github.com/zeldaret/oot/pull/814))
- The old `rgba5a1` format is still supported, but ZAPD will output a deprecation warning for every `rgb5a1` used.

### DIFF
--- a/ZAPD/ZTexture.cpp
+++ b/ZAPD/ZTexture.cpp
@@ -797,7 +797,7 @@ std::string ZTexture::GetExternalExtension() const
 	case TextureType::RGBA32bpp:
 		return "rgba32";
 	case TextureType::RGBA16bpp:
-		return "rgb5a1";
+		return "rgba16";
 	case TextureType::Grayscale4bpp:
 		return "i4";
 	case TextureType::Grayscale8bpp:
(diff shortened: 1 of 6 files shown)


fig02 (approved):

ZAPD/ZTexture.cpp
```diff
@@ -831,8 +831,19 @@ TextureType ZTexture::GetTextureTypeFromString(std::string str)
 
 	if (str == "rgba32")
 		texType = TextureType::RGBA32bpp;
+	else if (str == "rgba16")
+		texType = TextureType::RGBA16bpp;
 	else if (str == "rgb5a1")
+	{
 		texType = TextureType::RGBA16bpp;
+#ifndef DEPRECATION_OFF
```
NEstelami: Minor thing but I would prefer if we had ON defines instead of OFF. Double negatives can sometimes get confusing.

AngheloAlf: Done.

NEstelami (approved):

PR #144, opened 2021-05-23
A few Limbs, Skel and CurveAnimation fixes

AngheloAlf:
- Check if the `DList` of a `Limb` is pointing in the middle of another `DList`, and reuse that DL instead of re-extracting it.
- Check if the `DList` of a `Limb` is pointing past the object's size and don't try to extract it in that case.
  - Fixes crashing with a few unused objects.
- Fixes `Limb` names not being used if the `Limb` was declared before than a `Skeleton` in the XML.
- Don't try to extract the `SkelOffset` declared in a `CurveAnimation`.
  - Will simply look up the limb count value instead.
- Removes a few redundant pointers.


fig02:
unrelated to the changes here, though related to skeletons, i noticed recently that the formatting changed a little bit. imo some of the shorter entries benefit from being all on one line.
Was this change because of OutputFormatter?

heres an example of what i mean
```c
StandardLimb sAnubiceSkelLimb_0038A0 = { 
    { 0, 4700, 0 },
    0x01, 0xFF,
    NULL
}; 

StandardLimb sAnubiceSkelLimb_0038AC = { 
    { 0, 0, 0 },
    0x02, 0x0B,
    NULL
}; 
```
imo is better as
```c
StandardLimb sAnubiceSkelLimb_0038A0 = {  { 0, 4700, 0 }, 0x01, 0xFF, NULL }; 
StandardLimb sAnubiceSkelLimb_0038AC = {  { 0, 0, 0 }, 0x02, 0x0B, NULL }; 
```

AngheloAlf:
No, it isn't related to `OutputFormatter`.
It is being outputted like this because of limbs like this one
```c
StandardLimb sbject_ganon_Skel_0114E8Limb_0113F4 = {  { 316, 0, 0 }, 0xFF, 0xFF, object_ganon_DL_006460 }; 
```

I think I could add a check to change how the Limb's code is outputted depending if the DList is a `NULL` or it isn't.

fig02:
ah ok. yea these will also get shorter when they have proper names instead of auto generated ones.
not a big deal then, was just curious. i think most of these entries will eventually be able to fit on one line but no need to worry about it now

fig02 (approved):

Issue #17, opened 2020-12-29
ZAPD needs support for external segment references


Kenix3:
This would be where for example an object file references a display list from game_play keep. ZAPD should be able to manage this via a reference in the XML file for the object.

AngheloAlf:
This issue should be addressed now.

PR #151, opened 2021-06-05
`ExternalFile`: Expose other XML's declarations during extraction

AngheloAlf:
As we know, lots of objects cross-reference stuff from other objects (for example, `gameplay_keep` stuff). So this PR aims to provide a way to tell ZAPD where it should try to look to grab those symbols.

- Add `ExternalFile` node.
  - Allows to tell what extra XMLs files ZAPD should be looking for when looking for symbols.
  - Needs 2 attributes, `XmlPath` and `OutPath`.
    - `XmlPath`: Path to the XML to be read. Relative to the path defined in `ExternalXMLFolder`.
    - `OutPath`: Path to the folder where this object will be extracted to (This is used mainly for the `#include ".h"`).
- Add `ExternalFile` and `ExternalXMLFolder` to config file.
  - `ExternalFile` declared in the config file has the same effect as one declared in an object, but with the difference that this external file would be "global", so the declarations from that XML would be exposed to every file extracted (useful for `gameplay_keep` for example).
  - `ExternalXMLFolder` stores the path to the XMLs folder.
- Add method `GetSegmentedPtrName` to Globals.
  - Receives a segmented pointer and searches in every file for a symbol for that address.
- Changes in `ZResource`:
  - ~~(Almost) always declare as soon as possible every resource defined in the XML in `ExtractFromXML`.~~
  - ~~Add method `GetDeclarationAlignment`. Allows to define the alignment of said resource.~~
  - ~~Add method `GetDeclarationPadding`. Allows to define the padding of said resource.~~
  - ~~Add method `DeclareVar`. Allows to define the default way to declare the resource.~~
  - ~~Add method `GetDefaultName`. Returns a default name for a resource, for the cases when a resource is not defined in XML so an autogenerated name is needed.~~
  - ~~`GetSourceOutputCode` declares the resource in case it has not been declared yet.~~
- Cleanup in `ZAnimation`, ~~`ZVtx`~~, ~~`ZVector`~~, ~~`ZSkeleton`~~, `SetMesh`, ~~`ZPath`~~, ~~`ZMtx`~~, `ZArray`, among others.
- Refactor `ZCollision`.
- Refactor `ZLimb`.
- Remove extra constructors.
- Change warnings for invalid Skeleton types. Now an error will be triggered in those cases.
  - ~~This PR will probably fail because of this change.~~
- ~~Small changes in `AddDeclaration*` methods to prevent some memory leaks.~~
- ~~Throw error if `Segment` is missing in a `File` node.~~ Some XMLs don't have a clear segment, so this check doesn't make sense anymore.
- Properly extract `Vtx` arrays to a file.
  - For some reason, some of those arrays were being extracted to files, while others weren't.
- Convert some pointers to normal instances to prevent memory leaks and extra heaps allocations.
- ~~Extract cutscenes as `CutsceneData[]` instead of `s32[]`.~~ Fixed in another PR.
- Probably something else I missed.

Example of the new stuff:
Adding the following to the ZAPD config file allows to expose `gameplay_keep`'s symbols to every extracted file
```xml
    <ExternalXMLFolder Path="assets/xml/"/>
    <ExternalFile XmlPath="objects/gameplay_keep.xml" OutPath="objects/gameplay_keep/"/>
```

This way, the symbols from segment 4 will be used instead of the raw hex address. 
For example in `object_zo`, the following dlist would change from this
```c
Gfx gZoraRipplesVisualDL[] = {
    gsDPPipeSync(),
    gsDPSetCombineLERP(PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT, PRIMITIVE, 0, TEXEL0, 0, 0, 0, 0, COMBINED, 0, 0, 0,
                       COMBINED),
    gsDPSetRenderMode(G_RM_FOG_SHADE_A, G_RM_ZB_CLD_SURF2),
    gsSPTexture(0xFFFF, 0xFFFF, 0, G_TX_RENDERTILE, G_ON),
    gsDPLoadTextureBlock(0x040190B0, G_IM_FMT_I, G_IM_SIZ_8b, 64, 64, 0, G_TX_NOMIRROR | G_TX_WRAP, G_TX_NOMIRROR
                         | G_TX_WRAP, G_TX_NOMASK, G_TX_NOMASK, G_TX_NOLOD, G_TX_NOLOD),
    gsSPClearGeometryMode(G_CULL_BACK),
    gsSPEndDisplayList(),
}; 
```
To this
```c
Gfx gZoraRipplesVisualDL[] = {
    gsDPPipeSync(),
    gsDPSetCombineLERP(PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT, PRIMITIVE, 0, TEXEL0, 0, 0, 0, 0, COMBINED, 0, 0, 0,
                       COMBINED),
    gsDPSetRenderMode(G_RM_FOG_SHADE_A, G_RM_ZB_CLD_SURF2),
    gsSPTexture(0xFFFF, 0xFFFF, 0, G_TX_RENDERTILE, G_ON),
    gsDPLoadTextureBlock(gEffShockwaveTex, G_IM_FMT_I, G_IM_SIZ_8b, 64, 64, 0, G_TX_NOMIRROR | G_TX_WRAP, G_TX_NOMIRROR
                         | G_TX_WRAP, G_TX_NOMASK, G_TX_NOMASK, G_TX_NOLOD, G_TX_NOLOD),
    gsSPClearGeometryMode(G_CULL_BACK),
    gsSPEndDisplayList(),
}; 
```
(notice now it uses the texture instead of the raw symbol).

Related issue: #17


NEstelami (commented):
With regards to the `ExternalXMLFolder` node, do you think XMLs are ever not going to be stored in just `assets/xml`?

AngheloAlf:
`ExternalXMLFolder` allows to specify where we store the XMLs. If in any time we change were those XMLs are, we can change that value easily. The whole idea is to not hardcode it in the code.

NEstelami:
Makes sense. Also with `ExternalFile`, when two external XMLs that occupy the same segment are declared in both the config and the object XML, does the object XML's declaration take priority? There are certain cases where two or more objects occupy the same segment depending on context (ie. gameplay_field_keep and gameplay_dangeon_keep both occupying segment 5).

AngheloAlf:
In that case both files are considered and the first one in the list takes priority.
The system allows to check for types when looking for a symbol. 
For example, if you need a symbol from segment 5 at offset `0xCA70`with type `Gfx`, and both gameplay_field_keep and gameplay_dangeon_keep have an asset at that offset, but the former has a `Vtx` and the latter has a `Gfx`, then the symbol from `gameplay_dangeon_keep` will be used.
If both objects have an asset at the same offset, same segment and same type, then it would be a problem. But that has not happened yet in my tests. We could also not declare field_keep and dangeon_keep as global ExternalFiles (in the config file), but use them only if a specific object need one of them

NEstelami:
IIRC bosses like Morpha map certain assets to segment 8, where icon_item_static is normally loaded, and access them from offset 0 where there will probably be a conflict.

fig02:
8 - 12 (and maybe more) get used for many different purposes, its only item_icon_static in the context of z_parameter and pause menu really
but yes agreed that we will need to account for the dynamic purposes of all of these segments

AngheloAlf:
We can simply not declare those XMLs as global external files. Only the external files that are explicitely declared as such would be considered during extraction.

Using the example from the post:
```xml
    <ExternalXMLFolder Path="assets/xml/"/>
    <ExternalFile XmlPath="objects/gameplay_keep.xml" OutPath="objects/gameplay_keep/"/>
```
In this case, only `gameplay_keep` will be used as an external file.
So, to be specific, ZAPD will not scan every XML in the path specified by`ExternalXMLFolder`.

ZAPD/GameConfig.cpp
```diff
@@ -100,15 +106,42 @@ void GameConfig::ConfigFunc_BGConfig(const tinyxml2::XMLElement& element)
 	bgScreenHeight = element.IntAttribute("ScreenHeight", 240);
 }
 
-void GameConfig::ReadConfigFile(const std::string& argConfigFilePath)
+void GameConfig::ConfigFunc_ExternalXMLFolder(const tinyxml2::XMLElement& element)
+{
+	if (externalXmlFolder == "")
+		externalXmlFolder = element.Attribute("Path");
```
NEstelami: Should there be some kind of error check here if `Path` is not valid like there is with `ConfigFunc_ExternalFile`?

AngheloAlf: Yeah. I just added the error checks

NEstelami (approved):

PR #157, opened 2021-07-04
Fix DLists' IA4 format detection 

AngheloAlf:
Properly extract IA4 textures with their correct format, instead of I4.
This bug seems to only happen in textures that are not declared in the XML, but are referenced by a dlist.
One example of this issue is the only texture in `object_gi_butterfly`.


Zelllll (approved):
Looks good!

NEstelami (approved):

PR #164, opened 2021-08-13
Fix compilation with -Wextra

Random06457:
This PR:

- Adds `-Wextra -Werror` to the compilation flags and fixes all the errors associated.
- Marks `CRC32B` and `CRC32BCT` as `maybe_unused`
- Replaces `GfxdCallback_FormatSingleEntry(void)` with `GfxdCallback_FormatSingleEntry()` since this is not required in C++.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ COPYCHECK_ARGS ?=
 
 CXX := g++
 INC := -I ZAPD -I lib/assimp/include -I lib/elfio -I lib/json/include -I lib/stb -I lib/tinygltf -I lib/libgfxd -I lib/tinyxml2 -I ZAPDUtils
-CXXFLAGS += -fpic -std=c++17 -Wall -fno-omit-frame-pointer
+CXXFLAGS += -fpic -std=c++17 -Wall -Wextra -Werror -fno-omit-frame-pointer
 
 ifneq ($(DEBUG),0)
   OPTIMIZATION_ON = 0
--- a/ZAPD/ZLimb.cpp
+++ b/ZAPD/ZLimb.cpp
@@ -447,7 +447,6 @@ void ZLimb::ParseRawData()
 	case ZLimbType::LOD:
 		dList2Ptr = BitConverter::ToUInt32BE(rawData, rawDataIndex + 12);
 		[[fallthrough]];
-		// Intended fallthrough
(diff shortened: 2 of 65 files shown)


ZAPD/ZLimb.cpp
```diff
@@ -446,6 +446,7 @@ void ZLimb::ParseRawData()
 	{
 	case ZLimbType::LOD:
 		dList2Ptr = BitConverter::ToUInt32BE(rawData, rawDataIndex + 12);
+		[[fallthrough]];
 		// Intended fallthrough
```
AngheloAlf: That's interesting. You can remove the comment now.
```suggestion
		[[fallthrough]];
```

AngheloAlf (approved):

AngheloAlf:
I noted a problem with this PR. I think you shouldn't add `-Werror`
`clang++` emits a lot more of warnings than `g++`. This branch has some warnings with `clang++`.
Why is `clang++` important? Because Mac alias `clang++` as `g++`, so Mac users will not be able to build ZAPD anymore.
So, either you remove `-Werror` or fix those clang warnings.

Makefile
```diff
@@ -7,7 +7,7 @@ COPYCHECK_ARGS ?=
 
 CXX := g++
 INC := -I ZAPD -I lib/assimp/include -I lib/elfio -I lib/json/include -I lib/stb -I lib/tinygltf -I lib/libgfxd -I lib/tinyxml2 -I ZAPDUtils
-CXXFLAGS += -fpic -std=c++17 -Wall -fno-omit-frame-pointer
+CXXFLAGS += -fpic -std=c++17 -Wall -Wextra -Werror -fno-omit-frame-pointer
```
AngheloAlf, line 10: I was thinking about this. Could you change it so `-Werror` is not used in `DEBUG` builds? It could be a bit annoying :sweat_smile:

Random06457: I don't see why it would be annoying. The way I see it, warnings should be avoided regardless of the type of build.

AngheloAlf: Sometimes I want to make some quick tests that may produce some warnings while developing, so I can decide if I would really want to properly implement that or scrap the idea and move on.
I think I could adapt myself to -Werror anyways.

NEstelami (approved):

PR #166, opened 2021-08-20
Switches, some dictionaries, various other small bits of cleanup

EllipticEllipsis:
The main point of this PR is to convert most of the big if-else chains to switches or other devices, such as dictionary searches, that are either more efficient or easier to read. Along the way I fixed a few minor things like `NULL -> nullptr`, incorrect asserts, inserted a few more errors, removed some redundant wrapper function usage, etc. I also tweaked one setting in the formatter to make switches more readable, since there are even more now. It looks like I touched a lot of files, but a lot of them are just the formatter, I promise!

Considerable thanks must go to @AngheloAlf for helping me with learning about this codebase and C++; any glaring howlers are no doubt my fault, so please do point them out.


ZAPD/Globals.cpp
```diff
@@ -1,7 +1,7 @@
 #include "Globals.h"
-#include <algorithm>
 #include <Utils/File.h>
 #include <Utils/Path.h>
```
AngheloAlf: I know this wasn't you, but those includes aren't including a system library, so those should be using `""` instead of `<>`.
Since this is a clean-up PR, would be nice to change those too.

ZAPD/Main.cpp
```diff
@@ -35,18 +35,25 @@ void BuildAssetBlob(const fs::path& blobFilePath, const fs::path& outPath);
...
+		"\tGRUMBLE,GRUMBLE...",
+		"\tDODONGO DISLIKES SMOKE \n\tAnd ZAPD dislikes whatever you fed it.",
+		"\tMay the way of the Hero lead \n\tto the debugger."
+		"\tTHE WIND FISH SLUMBERS LONG... \n\tTHE HERO'S LIFE GONE... "
+		"\tSEA BEARS FOAM, SLEEP BEARS DREAMS. \n\tBOTH END IN THE SAME WAY CRASSSH!"
 		"\tYou've met with a terrible fate, haven't you?",
-		"\tSEA BEARS FOAM. SLEEP BEARS DREAMS. \n\tBOTH END IN THE SAME WAY: CRASSSH!",
-	};
+		"\tMaster, I calculate a 100% probability that ZAPD has crashed. \n\tAdditionally, the "
+		"batteries in your Wii Remote are nearly depleted."};
```
Random06457: This is a very important change, to which I would suggest adding the now famous:
```
    CONGRATURATIONS!    
All Pages are displayed.
       THANK YOU!       
 You are great debugger!
 ```

File: ZAPD/Main.cpp [hunk 3]
```diff
@@ -35,18 +35,25 @@ void BuildAssetBlob(const fs::path& blobFilePath, const fs::path& outPath);
...
 
 	// Feel free to add more crash messages.
 	const char* crashEasterEgg[] = {
+		"\tIT'S A SECRET TO EVERYBODY. \n\tBut it shouldn't be, you'd better ask about it!",
+		"\tI AM ERROR.",
+		"\tGRUMBLE,GRUMBLE...",
+		"\tDODONGO DISLIKES SMOKE \n\tAnd ZAPD dislikes whatever you fed it.",
+		"\tMay the way of the Hero lead \n\tto the debugger."
+		"\tTHE WIND FISH SLUMBERS LONG... \n\tTHE HERO'S LIFE GONE... "
+		"\tSEA BEARS FOAM, SLEEP BEARS DREAMS. \n\tBOTH END IN THE SAME WAY CRASSSH!"
```
AngheloAlf: I think you are missing some commas there
```suggestion
		"\tDODONGO DISLIKES SMOKE \n\tAnd ZAPD dislikes whatever you fed it.",
		"\tMay the way of the Hero lead \n\tto the debugger.",
		"\tTHE WIND FISH SLUMBERS LONG... \n\tTHE HERO'S LIFE GONE... ",
		"\tSEA BEARS FOAM, SLEEP BEARS DREAMS. \n\tBOTH END IN THE SAME WAY CRASSSH!",
```

AngheloAlf (changes requested):

ZAPD/Main.cpp
EllipticEllipsis: Yeah, the formatter doesn't make it easy with some of the longer ones...

ZAPD/Globals.cpp
```diff
@@ -66,6 +66,22 @@ std::string Globals::FindSymbolSegRef(int32_t segNumber, uint32_t symbolAddress)
...
+{
+	SYMBOL_MAP,
+	SEGMENT,
+	ACTOR_LIST,
+	OBJECT_LIST,
+	TEXTURE_POOL,
+	BG_CONFIG
+};
+
+const static std::map<std::string, ConfigType> ConfigTypeDictionary = {
```
leoetlino: While `const static` is valid C++, `static const` is the usual order in which those two declaration specifiers are written. (And C goes even further by deprecating the use of storage-class specifiers other than at the beginning of a declaration.)

`const static` is valid, but it looks weird to anybody who's familiar with C or C++ (kind of like how there is a specific order for adjectives in English...) :)

leoetlino: `<map>` should be included, otherwise this is relying on an indirect include and will fail to compile if that include disappears for whatever reason in the future.

ZAPD/Globals.cpp
```diff
@@ -82,47 +98,67 @@ void Globals::ReadConfigFile(const std::string& configFilePath)
...
 	     child = child->NextSiblingElement())
 	{
-		if (std::string(child->Name()) == "SymbolMap")
-		{
-			std::string fileName = std::string(child->Attribute("File"));
-			GenSymbolMap(Path::GetDirectoryName(configFilePath) + "/" + fileName);
-		}
-		else if (std::string(child->Name()) == "Segment")
+		auto it = ConfigTypeDictionary.find(child->Name());
+		if (it != ConfigTypeDictionary.end())
```
leoetlino: IMO it would be better to put the failure branch first and `continue`, so as to reduce the indentation level for the rest of the code.

ZAPD/Globals.cpp
```diff
@@ -82,47 +98,67 @@ void Globals::ReadConfigFile(const std::string& configFilePath)
...
-			std::string fileName = std::string(child->Attribute("File"));
-			ReadTexturePool(Path::GetDirectoryName(configFilePath) + "/" + fileName);
+				case ConfigType::ACTOR_LIST:
+				{
+					std::string fileName = child->Attribute("File");
+					std::vector<std::string> lines =
+						File::ReadAllLines(Path::GetDirectoryName(configFilePath) + "/" + fileName);
+
+					for (std::string line : lines)
+						cfg.actorList.push_back(line);
```
leoetlino: Unlikely to matter in practice, but note that this performs a double copy (`line` is copied from the vector, and then a second copy occurs when it is inserted into cfg.actorList). You can eliminate both copies (and allocations) by doing:

```cpp
for (auto& line : lines)
  cfg.actorList.emplace_back(std::move(line));
```

ZAPD/Globals.cpp
```diff
@@ -82,47 +98,67 @@ void Globals::ReadConfigFile(const std::string& configFilePath)
...
+					cfg.bgScreenHeight = child->IntAttribute("ScreenHeight", 240);
+					break;
+			}
 		}
-		else if (std::string(child->Name()) == "BGConfig")
+		else
 		{
-			cfg.bgScreenWidth = child->IntAttribute("ScreenWidth", 320);
-			cfg.bgScreenHeight = child->IntAttribute("ScreenHeight", 240);
+			fprintf(stderr, "Unsupported configuration variable: %s", child->Name());
```
leoetlino:
```suggestion
			fprintf(stderr, "Unsupported configuration variable: %s\n", child->Name());
```

ZAPD/Main.cpp
```diff
@@ -35,18 +35,29 @@ void BuildAssetBlob(const fs::path& blobFilePath, const fs::path& outPath);
 #define ARRAY_COUNT(arr) (sizeof(arr) / sizeof(arr[0]))
 void ErrorHandler(int sig)
 {
-	void* array[4096];
-	const size_t nMaxFrames = sizeof(array) / sizeof(array[0]);
+	void* array[0x1000];
+	const size_t nMaxFrames = ARRAY_COUNT(array);
```
leoetlino: This can just be
```suggestion
	const size_t nMaxFrames = std::size(array);
```
No need for ugly macros.

ZAPD/Overlays/ZOverlay.cpp
```diff
@@ -222,15 +222,23 @@ std::string ZOverlay::GetSourceOutputCode(const std::string& prefix)
...
-		return SectionType::Text;
-	else if (sectionName == ".rel.data" || sectionName == ".data")
-		return SectionType::Data;
-	else if (sectionName == ".rel.rodata" || sectionName == ".rodata" ||
-	         sectionName == ".rodata.str1.4" || sectionName == ".rodata.cst4")
-		return SectionType::RoData;
-	else if (sectionName == ".rel.bss" || sectionName == ".bss")
-		return SectionType::Bss;
-
+	const static std::map<std::string, SectionType> SectionTypeDictionary = {
```
leoetlino: static const

ZAPD/ZDisplayList.cpp
```diff
@@ -2125,45 +2125,37 @@ bool ZDisplayList::TextureGenCheck(ZRoom* scene, ZFile* parent, std::string pref
...
-	{
-		if (siz == F3DZEXTexSizes::G_IM_SIZ_4b)
-			return TextureType::GrayscaleAlpha4bpp;
-		else if (siz == F3DZEXTexSizes::G_IM_SIZ_8b)
-			return TextureType::GrayscaleAlpha8bpp;
-		else if (siz == F3DZEXTexSizes::G_IM_SIZ_16b)
-			return TextureType::GrayscaleAlpha16bpp;
-	}
-	else if (fmt == F3DZEXTexFormats::G_IM_FMT_I)
+	const static std::map<std::tuple<F3DZEXTexFormats, F3DZEXTexSizes>, TextureType>
```
leoetlino: static const

ZAPD/Overlays/ZOverlay.cpp
leoetlino: `#include <map>`

ZAPD/ZDisplayList.cpp
leoetlino: `#include <map>`, `#include <tuple>`

ZAPD/ZFile.cpp
```diff
@@ -71,6 +71,13 @@ ZFile::~ZFile()
 	}
 }
 
+const static std::map<std::string, ZGame> ZGameDictionary = {
```
leoetlino: static const + map include

ZAPD/ZFile.cpp
```diff
@@ -86,27 +93,22 @@ void ZFile::ParseXML(ZFileMode mode, XMLElement* reader, std::string filename, b
...
-		if (std::string(gameStr) == "MM")
-			Globals::Instance->game = ZGame::MM_RETAIL;
-		else if (std::string(gameStr) == "SW97" || std::string(gameStr) == "OOTSW97")
-			Globals::Instance->game = ZGame::OOT_SW97;
-		else if (std::string(gameStr) == "OOT")
-			Globals::Instance->game = ZGame::OOT_RETAIL;
-		else
-			throw std::runtime_error(
-				StringHelper::Sprintf("Error: Game type %s not supported.", gameStr));
+	if (gameStr != nullptr) {
```
leoetlino: brace style?

ZAPD/ZLimb.cpp
```diff
@@ -618,47 +618,37 @@ const char* ZLimb::GetSourceTypeName(ZLimbType limbType)
...
-	if (attrName == "Skin")
-	{
-		return ZLimbType::Skin;
-	}
-	if (attrName == "Curve")
-	{
-		return ZLimbType::Curve;
-	}
-	if (attrName == "Legacy")
+	const static std::map<std::string, ZLimbType> ZLimbTypeDictionary = {
```
leoetlino: static const + include

ZAPD/ZScalar.cpp
```diff
@@ -28,129 +28,107 @@ void ZScalar::ParseXML(tinyxml2::XMLElement* reader)
...
-	else if (type == "x64")
-	{
-		return ZScalarType::ZSCALAR_X64;
-	}
-	else if (type == "f32")
-	{
-		return ZScalarType::ZSCALAR_F32;
-	}
-	else if (type == "f64")
+	const static std::map<std::string, ZScalarType> ZScalarTypeDictionary = {
```
leoetlino: static const + include

ZAPD/ZSkeleton.cpp
```diff
@@ -160,6 +157,22 @@ ZResourceType ZSkeleton::GetResourceType() const
 	return ZResourceType::Skeleton;
 }
 
+ZSkeletonType ZSkeleton::GetTypeByAttributeName(const std::string& attrName)
+{
+	const static std::map<std::string, ZSkeletonType> ZSkeletonTypeDictionary = {
```
leoetlino: static const + include

ZAPD/ZTexture.cpp
```diff
@@ -820,54 +822,35 @@ fs::path ZTexture::GetPoolOutPath(const fs::path& defaultValue)
 	return defaultValue;
 }
 
-TextureType ZTexture::GetTextureTypeFromString(std::string str)
+TextureType ZTexture::GetTextureTypeFromString(std::string type)
 {
-	TextureType texType = TextureType::Error;
+	const static std::map<std::string, TextureType> ZTextureTypeDictionary = {
```
leoetlino: static const + include

ZAPD/ZDisplayList.cpp
```diff
@@ -2125,45 +2125,37 @@ bool ZDisplayList::TextureGenCheck(ZRoom* scene, ZFile* parent, std::string pref
...
-	if (fmt == F3DZEXTexFormats::G_IM_FMT_RGBA)
-	{
-		if (siz == F3DZEXTexSizes::G_IM_SIZ_16b)
-			return TextureType::RGBA16bpp;
-		else if (siz == F3DZEXTexSizes::G_IM_SIZ_32b)
-			return TextureType::RGBA32bpp;
-	}
-	else if (fmt == F3DZEXTexFormats::G_IM_FMT_CI)
-	{
-		if (Globals::Instance->useLegacyZDList)
```
leoetlino, line 2050: Was the removal of this check intended?

ZAPD/Globals.cpp
EllipticEllipsis: Oh, I didn't think it mattered... I thought these were the same type of thing: to me `const` just means "put in rodata" and `static` means "only usable in this file", or "put it in data and not on the stack", which sound like the same category to me.

EllipticEllipsis: And yeah, I see what you mean with the include... I'm used to decomp where basically everything's included in every file.

ZAPDUtils/Utils/MemoryStream.cpp
```diff
@@ -30,12 +30,19 @@ uint64_t MemoryStream::GetLength()
...
+			break;
+
+		case SeekOffsetType::Current:
+			baseAddress += offset;
+			break;
+
+		case SeekOffsetType::End:
+			baseAddress = bufferSize - offset - 1;
+			break;
+	}
```
NEstelami: I understand using switch statements as replacements for long if else chains, but it seems there are instances where the switch statements become more bloated than the if-else chains they replace.

ZAPD/Overlays/ZOverlay.cpp
```diff
@@ -222,15 +222,23 @@ std::string ZOverlay::GetSourceOutputCode(const std::string& prefix)
...
+		{".rodata.cst4", SectionType::RoData},
+		{".bss", SectionType::Bss},
+		{".rel.bss", SectionType::Bss},
+	};
+
+	auto it = SectionTypeDictionary.find(sectionName);
+	if (it != SectionTypeDictionary.end())
+	{
+		return it->second;
+	}
```
NEstelami: Here too.

ZAPD/Globals.cpp
```diff
@@ -66,6 +66,22 @@ std::string Globals::FindSymbolSegRef(int32_t segNumber, uint32_t symbolAddress)
...
+
+const static std::map<std::string, ConfigType> ConfigTypeDictionary = {
+	{"SymbolMap", ConfigType::SYMBOL_MAP},     {"Segment", ConfigType::SEGMENT},
+	{"ActorList", ConfigType::ACTOR_LIST},     {"ObjectList", ConfigType::OBJECT_LIST},
+	{"TexturePool", ConfigType::TEXTURE_POOL}, {"BGConfig", ConfigType::BG_CONFIG},
+};
+
 void Globals::ReadConfigFile(const std::string& configFilePath)
 {
 	tinyxml2::XMLDocument doc;
```
NEstelami: Changing from an if-else chain to a switch statement here doesn't seem to offer much benefit. You're still left with a chain (one of cases), and with the added enum and dictionary it seems it's gotten bulkier.

ZAPD/Globals.cpp
EllipticEllipsis: Oh, this looks much better, thanks. Is there a general rule about if error cases go first or last?

leoetlino: I don't have an answer for ZAPD (you'd have to ask @NEstelami) but early returns/exits are common in many other projects (e.g. [LLVM](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code))

ZAPD/ZDisplayList.cpp
EllipticEllipsis: Nope, thanks for catching that 😓

EllipticEllipsis:
I would say there are 6 main advantages to using dictionaries and switches
- Switches are efficient: O(1) instead of O(n) for if-elses, while `std::map` is designed to be O(log n) for lookups. Even for small switches we gain from the compiler being able to choose to optimise more efficiently.
- With dictionary lookup, the code correctly reflects the uniformity of the input: I do not have to read through every case to make sure they are treated the same way: the lookup guarantees that they are.
- If one switch case check is slightly wrong, it generally doesn't affect the others sigficantly. If one of the ifs is wrong, the whole chain after it becomes unreliable.
- If I want to add a case, I can just add it to the enum and dictionary: I do not have to add an entire extra branch. This ensures continued uniformity of code: accidents are far less likely. Given the direction we are moving in on configuration files, we'll want to be able to do this in future. This also marries nicely with switch structure: it's much better than comparing strings in ifs all the time.
- Switches enable a more complex fallthrough structure, should we later wish to take advantage of this.
- I'd also take readability over any putative bloat any day, and I think these switches are simply much more readable. How often in OoT do we lament the lack of switches in Player?

NEstelami:
> I would say there are 6 main advantages to using dictionaries and switches
> 
> * Switches are efficient: O(1) instead of O(n) for if-elses, while `std::map` is designed to be O(log n) for lookups. Even for small switches we gain from the compiler being able to choose to optimise more efficiently.
> * With dictionary lookup, the code correctly reflects the uniformity of the input: I do not have to read through every case to make sure they are treated the same way: the lookup guarantees that they are.
> * If one switch case check is slightly wrong, it generally doesn't affect the others sigficantly. If one of the ifs is wrong, the whole chain after it becomes unreliable.
> * If I want to add a case, I can just add it to the enum and dictionary: I do not have to add an entire extra branch. This ensures continued uniformity of code: accidents are far less likely. Given the direction we are moving in on configuration files, we'll want to be able to do this in future. This also marries nicely with switch structure: it's much better than comparing strings in ifs all the time.
> * Switches enable a more complex fallthrough structure, should we later wish to take advantage of this.
> * I'd also take readability over any putative bloat any day, and I think these switches are simply much more readable. How often in OoT do we lament the lack of switches in Player?

- It is nice that you are thinking about performance but in a scenario like this I don't think it will make a noticeable difference. It's generally not good practice to try to optimize something you haven't proved to be performance intensive - that just tends to waste time without yielding much of any benefit.
- Not sure I understand what you mean by your second point.
- I do not think that one incorrect if statement would necessarily ruin the entire chain any more than a switch statement would. If an incorrect switch case makes your code break early, it's going to have a similar effect to an incorrect if that is followed by an else.
- The enum and dictionary seems to be adding an extra abstraction layer without making the code cleaner. You still have all your config options in one giant block inside a function. That does not change. If you're thinking about the implications of having many different config options, I would take a look at how the `REGISTER_ZFILENODE` macro is used. That macro allowed us to eliminate a rather large if-else/switch chain for every possible Resource that could be inside an XML.
- Having fallthroughs can be nice, but doesn't seem to be utilized in these instances.
- I'm not sure that in the instances I highlighted the if-else statements were hard to read or bloated. As I pointed out, in some of these cases the switch statements are more bloated than their if-else counterparts. I do not think switch statements are always better than if-else statements. They're a tool designed for a certain purpose, and like any tool I would argue they should be used where they shine best, not as permanent replacement to if statements.

leoetlino:
> * Switches are efficient: O(1) instead of O(n) for if-elses, while `std::map` is designed to be O(log n) for lookups. Even for small switches we gain from the compiler being able to choose to optimise more efficiently.

FWIW, modern compilers will typically switch between if-else and switches automatically and use the fastest option regardless of whether switches or if chains were used in the source code. (LLVM can even materialize lookup tables out of functions that look like `switch (foo) { case 0: return x; case 1: return y; case 2: return z; ... }`).

In my opinion, the real benefit with using switches (other than the readability improvement) is that the compiler can check whether the switch cases are exhaustive. If you add a new enum case and forget to update a switch statement, the compiler will warn you about it.

ZAPDUtils/Utils/MemoryStream.cpp
EllipticEllipsis: This one I concede doesn't offer much of an improvement, but since the enum's there anyway, it seemed worth it to me. I don't mind reverting this one.

ZAPD/Overlays/ZOverlay.cpp
EllipticEllipsis: I don't think this is any more difficult to read than the if-else: the rodata case in particular is a nuisance, what with having to look at the four different cases.

ZAPD/Globals.cpp
EllipticEllipsis: Addressed in #168 , so I'll revert this.

ZAPD/ZTexture.cpp
```diff
@@ -820,54 +823,33 @@ fs::path ZTexture::GetPoolOutPath(const fs::path& defaultValue)
 	return defaultValue;
 }
 
-TextureType ZTexture::GetTextureTypeFromString(std::string str)
+TextureType ZTexture::GetTextureTypeFromString(std::string type)
 {
-	TextureType texType = TextureType::Error;
+	static const std::map<std::string, TextureType> ZTextureTypeDictionary = {
```
AngheloAlf: I _think_ `unsorted_map` is faster for this kind of thing compared to plain `map`
From cppreference:
> Unordered associative containers implement unsorted (hashed) data structures that can be quickly searched (O(1) amortized, O(n) worst-case complexity). 

<https://en.cppreference.com/w/cpp/container>

EllipticEllipsis: I changed this where possible. Unfortunately `std::hash` can't handle fancy stuff like pairs or tuples without writing a whole new function to hash, so I couldn't change the ZDisplayListTexture one.

ZAPD/Globals.cpp
Kelebek1: Yeah, it's the standard as I know it too, decomp projects are an outlier in enforcing else/ifs everywhere. I think it makes the flow much more complicated and creates so much indenting and bracketing, I wish we could switch over to using returns.

NEstelami:
While there is a lot of good material in this PR, it looks like it won't be able to be finished given what has been expressed previously. Unfortunately it will have to be closed. Thank you for your hard work Elliptic, hopefully what's here can be worked into a future PR.

PR #168, opened 2021-08-21
Divide ReadConfigFile into separate functions and move them and the rest of GameConfig to their own file

EllipticEllipsis:
Following conversation in Discord, split the configuration file-handling function off into its own file and separated it into one function per config option, all within the `GameConfig` class.

I'm not sure if it's worth keeping `ReadTexturePool` and `GenSymbolMap` separate any more, would welcome suggestions.

Thanks to @AngheloAlf and @leoetlino for help with converting my classfree version to something that vaguely resembles C++, and @Jack-Walker for the original idea.

This is another PR where most of the changes are the formatter: only 6 files contain non-formatter changes.


ZAPD/GameConfig.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+	tinyxml2::XMLNode* root = doc.FirstChild();
+
+	if (root == nullptr)
+		return;
+
+	for (tinyxml2::XMLElement* child = root->FirstChildElement(); child != nullptr;
+	     child = child->NextSiblingElement())
+	{
+		if (std::string(child->Name()) == "Texture")
```
leoetlino: Probably doesn't matter, but you can avoid a memory allocation + a string copy by using a string_view instead of a string. (string_view will still let you compare strings with ==) 

```suggestion
		if (std::string_view(child->Name()) == "Texture")
```

(edit: `#include <string_view>` for std::string_view)

ZAPD/GameConfig.cpp
```diff
@@ -0,0 +1,138 @@
...
+		{
+			std::string crcStr = std::string(child->Attribute("CRC"));
```
leoetlino:
```suggestion
			std::string crcStr = child->Attribute("CRC");
```
or if you prefer to make the const char* -> std::string conversion explicit, I'd go with `auto crcStr = std::string(child->Attribute("CRC"));` (the "don't repeat yourself" principle)

ZAPD/GameConfig.cpp
```diff
@@ -0,0 +1,138 @@
...
+			fs::path texPath = std::string(child->Attribute("Path"));
+			std::string texName = "";
```
leoetlino:
```suggestion
			std::string texName;
```

ZAPD/GameConfig.cpp
```diff
@@ -0,0 +1,138 @@
...
+	};
+
+	configFilePath = argConfigFilePath;
+	tinyxml2::XMLDocument doc;
+	tinyxml2::XMLError eResult = doc.LoadFile(configFilePath.c_str());
+
+	if (eResult != tinyxml2::XML_SUCCESS)
+	{
+		throw std::runtime_error("Error: Unable to read config file.");
+		return;
```
leoetlino: return after throw is unnecessary.

ZAPD/GameConfig.h
```diff
@@ -0,0 +1,43 @@
...
+	fs::path path = "";  // Path to Shared Texture
+};
+
+class ZFile;
+
+class GameConfig
+{
+public:
+	std::string configFilePath;
+	std::map<int32_t, std::string> segmentRefs;
```
leoetlino, line 21: `<string>` and `<vector>` should probably be included too. And `<cstdint>` too for int32_t and uint32_t.

ZAPD/GameConfig.h
```diff
@@ -0,0 +1,43 @@
...
+	void ConfigFunc_Segment(const tinyxml2::XMLElement& element);
+	void ConfigFunc_ActorList(const tinyxml2::XMLElement& element);
+	void ConfigFunc_ObjectList(const tinyxml2::XMLElement& element);
+	void ConfigFunc_TexturePool(const tinyxml2::XMLElement& element);
+	void ConfigFunc_BGConfig(const tinyxml2::XMLElement& element);
+
+	void ReadConfigFile(const std::string& configFilePath);
+};
+
+typedef void (GameConfig::*ConfigFunc)(const tinyxml2::XMLElement&);
```
leoetlino:
```suggestion
using ConfigFunc = void (GameConfig::*)(const tinyxml2::XMLElement&);
```
but since this is not used anywhere else it might be better to just remove this from the header (and move it to the .cpp)

ZAPD/GameConfig.cpp
```diff
@@ -0,0 +1,138 @@
...
+			texturePool[crc].path = texPath;
+		}
+	}
+}
+
+void GameConfig::GenSymbolMap(const std::string& symbolMapPath)
+{
+	auto symbolLines = File::ReadAllLines(symbolMapPath);
+
+	for (std::string symbolLine : symbolLines)
```
leoetlino: I realise this is pre-existing code, but you can avoid a useless copy by taking a reference.
```suggestion
	for (std::string& symbolLine : symbolLines)
```

ZAPD/GameConfig.cpp
```diff
@@ -0,0 +1,138 @@
...
+	{
+		auto split = StringHelper::Split(symbolLine, " ");
+		uint32_t addr = strtoul(split[0].c_str(), nullptr, 16);
+		std::string symbolName = split[1];
+
+		symbolMap[addr] = symbolName;
```
leoetlino:
```suggestion
		symbolMap[addr] = std::move(symbolName);
```
would allow the symbolName to be moved into the map without any extra allocation or copy.

ZAPD/GameConfig.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+	std::string fileName = element.Attribute("File");
+	GenSymbolMap(Path::GetDirectoryName(configFilePath) + "/" + fileName);
+}
+
+void GameConfig::ConfigFunc_Segment(const tinyxml2::XMLElement& element)
+{
+	std::string fileName = element.Attribute("File");
+	int32_t segNumber = element.IntAttribute("Number");
+	segmentRefs[segNumber] = fileName;
```
leoetlino:
```suggestion
	segmentRefs[segNumber] = std::move(fileName);
```

ZAPD/GameConfig.cpp
EllipticEllipsis: Oh, this is neat. I think there's probably a lot of other places that could use this... well a lot that I didn't already turn into switches, anyway.

ZAPD/GameConfig.cpp
EllipticEllipsis: I worry about having an uninitialised variable floating around... is that safe? Or does `std::string` just assume this value automatically?

leoetlino: std::string's default constructor constructs an empty string so this is safe.

leoetlino (approved):

ZAPD/ZRoom/ZRoom.cpp
```diff
@@ -66,7 +67,7 @@ void ZRoom::ExtractFromXML(tinyxml2::XMLElement* reader, uint32_t nRawDataIndex)
 
 	scene = Globals::Instance->lastScene;
 
-	if (std::string(reader->Name()) == "Scene")
+	if (std::string_view(reader->Name()) == "Scene")
```
AngheloAlf: I don't really know how `string_view` works, but I don't think this the intended use.

EllipticEllipsis: @leoetlino suggested it: https://github.com/zeldaret/ZAPD/pull/168#discussion_r693412135

AngheloAlf: Oh, I didn't saw that.

leoetlino: It's basically a more convenient way to write `strcmp(reader->Name(), "Scene") == 0` (but clearer and possibly more optimised): https://godbolt.org/z/4Tzn861Yc

Comparing a std::string_view just requires figuring out the length of `reader->Name()` and then comparing against constant values.

AngheloAlf (approved):

NEstelami (approved):

PR #175, opened 2021-08-29
Show `rawDataSize` when `GetDListLength` errors

Dragorn421:
Would have made figuring out a dumb issue I ran across faster.

The issue was using the wrong file in the xml. Namely, `<File Name="testcube_flat.xml" Segment="6">` instead of `<File Name="testcube_flat.zobj" Segment="6">`.

### DIFF
--- a/ZAPD/ZDisplayList.cpp
+++ b/ZAPD/ZDisplayList.cpp
@@ -426,7 +426,7 @@ int32_t ZDisplayList::GetDListLength(const std::vector<uint8_t>& rawData, uint32
 				StringHelper::Sprintf("%s: Fatal error.\n"
 			                          "\t End of file found when trying to find the end of the "
 			                          "DisplayList at offset: '0x%X'.\n",
-			                          "Raw data size: 0x%X.\n",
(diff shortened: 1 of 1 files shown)


ZAPD/ZDisplayList.cpp
```diff
@@ -426,7 +426,8 @@ int32_t ZDisplayList::GetDListLength(const std::vector<uint8_t>& rawData, uint32
 				StringHelper::Sprintf("%s: Fatal error.\n"
 			                          "\t End of file found when trying to find the end of the "
 			                          "DisplayList at offset: '0x%X'.\n",
-			                          __PRETTY_FUNCTION__, rawDataIndex));
+			                          "Raw data size: 0x%X.\n",
```
leoetlino:
```suggestion
			                          "Raw data size: 0x%zX.\n",
```
Otherwise this is undefined behaviour since `rawDataSize` is a `size_t`.

leoetlino (changes requested):

AngheloAlf (approved):

Issue #61, opened 2021-01-06
Refactor ZResource::GetSourceCodeOutput to not return a string


NEstelami:
Returning a string is a legacy feature back before a declaration system was in place. The function would return the string with the C source code, but this is (for the most part) no longer the case, so we shouldn't be returning a string anymore.

PR #178, opened 2021-09-04
ZResource cleanup

AngheloAlf:
This PR was originally part of #151, but I decided to take some parts of it and PR them separately so it should be easier to review.

- Changes in `ZResource`:
  - (Almost) always declare as soon as possible every resource defined in the XML in `ExtractFromXML`.
  - Add method `GetDeclarationAlignment`. Allows to define the alignment of said resource.
  - Add method `GetDeclarationPadding`. Allows to define the padding of said resource.
  - Add method `DeclareVar`. Allows to define the default way to declare the resource.
  - Add method `GetDefaultName`. Returns a default name for a resource, for the cases 
  - when a resource is not defined in XML so an autogenerated name is needed.
- Add some brief docs to some methods of `ZResource`.
- Remove extra constructors.
- Add `offset_t` type.
  - Since we have `segptr_t` I though having this type would make sense.
  - Question: Should we drop the `_t` suffix because of UNIX compliance?
- Removed every redundant `ExtractFromXML` override, because most of them did essentially the same as the `ZResource` version.
- Reduce the use of the `GetSourceOutputCode` method. This way refactoring it will be simpler in the future (related to issue #61)

The idea is to minimize repeated code in every ZResource in the codebase.


NEstelami (approved):
Fantastic work on this PR! I honestly don't have too much to comment on it, everything here seems pretty solid. 👍 

Regarding your question on dropping the `_t`, considering it's already being used in a lot of places for things such as `uint32_t`, `int16_t`, etc, I think it can be kept.

Issue #143, opened 2021-05-18
Allow to declare AnimatedTextures/Materials in XMLs


AngheloAlf:
ZAPD currently is able to detect those in scenes, but they can't be declared in an XML file for objects.

One example is `object_death` at offsets `0xCB84` and `0xCBC0`.

AngheloAlf:
I could work in this after #140 is merged, just to prevent conflicts.

mzxrules:
what does it mean to be an animated texture?

AngheloAlf:
I'm not completely sure tbh. It is a MM asset type, so you may want to ask them instead.

PR #180, opened 2021-09-09
Texture Animation extraction from objects

EllipticEllipsis:
I completely rewrote ZAPD's `AnimatedMaterial` extraction to work separately from `ZRoom`. I have introduced two `ZResource`s to do this, one that handles the main array, and one that acts as a parent for the actual params data. These three in turn may point to other arrays, all of which are handled by ordinary structs. A diagram detailing the classes' inheritance may be found here: https://hackmd.io/@EllipticEllipsis/Hks1u5a-t, if there's better software for drawing such things, I'd be interested to know about it.

`SetAnimatedMaterialList` still exists, but has been rewritten to use the `ZResource`.

I introduced a number of warnings and errors to handle various possibilities that will cause extraction to fail or be incomplete: the most notable of these for the end-user is the one for undeclared textures in the Cycle texture array: previously, these were set using a hardcoded symbol name designed to match the one declared by the DLists, but this is not a practical long-term solution and **cannot** handle renaming: instead I opted to warn, giving the addresses of the textures so that the user can declare them manually for the extractor to find.

**N.B.** MM will need to declare the appropriate textures in rooms/scenes when mergeing this to avoid generating new warnings.

(One may ask why we cannot simply find and declare the textures automatically: the reason is that this system contains *no* information at all about the textures' types or dimensions. This information could in theory be obtained by scouring the DLists in the file for the appropriate segments, but this is not really practicable, particularly in the case of scenes/rooms where the DLists can be in an entirely different file; there are also likely edge cases where a segment is mentioned more than once in a file with different intended purposes, which we would have to choose between.)

I wrote some fairly extensive documentation on what the data is that this new file extracts and some details about how extraction is done; I think that ZAPD could do with a lot more of this where appropriate, since otherwise it is completely opaque what it is doing.

Addresses #143. This should clear the final obstacle to MM being able to do object decompilation.

Huge thanks are due to @AngheloAlf for guiding me through ZAPD's labyrinthine and mostly-undocumented innards throughout this work: it took a couple of weeks to do, and it would have easily taken 10 times as long without his help, if I would have been able to do it at all. Any flaws, bad coding etc. are entirely my own.


EllipticEllipsis:
I forgot to mention: this is what my new warnings/errors look like. I think making them stick out like this is much more helpful to the end user trying to find what's wrong. 
image

Format is not final: some of it (like the function name) should maybe be suppressed unless debugging or higher than default verbosity is on, but I think this is a good baseline.

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+ *
+ * Types
+ * ===
+ * There are 7 values that `T` can take:
+ *     `0`: A single texture scroll (Implemented by Gfx_TexScroll): subsidiary params are given as
+ *         an array with one entry, a struct `XX YY WW HH` containing xStep, yStep, width, height
+ *         (all u8).
+ *
+ *     `1`: A dual texture scroll (Implementated by Gfx_TwoTexScroll): same as type `0`, but with
+ *         two entries in the params array.
```
AngheloAlf, line 28: Could you eventually replicate all this documentation in the MM repo?

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+	for (int i = 0; i < count; i++)
+	{
+		bodyStr += StringHelper::Sprintf("    { %d, %d, 0x%02X, 0x%02X },\n", rows[i].xStep,
+		                                 rows[i].yStep, rows[i].width, rows[i].height);
+	}
+
+	bodyStr.pop_back();
+
+	// printf("bodyStr = %s\n", bodyStr.c_str());
```
AngheloAlf: Please remove commented code

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+			"\033[97m"
+			"TextureColorChangingParams::ParseRawData:"
+			"\033[0m"
+			"\033[91m"
+			" error: "
+			"\033[0m"
+			"\033[97m"
+			"color list length cannot be 0\n"
+			"\033[0m",
+			Globals::Instance->inputPath.c_str(), parent->GetName().c_str(), rawDataIndex));
```
AngheloAlf, line 255: ouch

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+	envColorListAddress = BitConverter::ToUInt32BE(rawData, rawDataIndex + 8);
+	frameDataListAddress = BitConverter::ToUInt32BE(rawData, rawDataIndex + 0xC);
+
+	uint32_t primColorListOffset = Seg2Filespace(primColorListAddress, parent->baseAddress);
+	uint32_t envColorListOffset = Seg2Filespace(envColorListAddress, parent->baseAddress);
+	uint32_t frameDataListOffset = Seg2Filespace(frameDataListAddress, parent->baseAddress);
+
+	uint32_t currentPtr;
+
+	// printf("cycleLength: %d\n", listLength);
```
AngheloAlf: Remove commented code

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+	uint32_t textureListOffset = Seg2Filespace(textureListAddress, parent->baseAddress);
+	uint32_t textureIndexListOffset = Seg2Filespace(textureIndexListAddress, parent->baseAddress);
+
+	uint32_t currentPtr;
+
+	uint8_t currentIndex;
+	uint8_t maxIndex = 0;  // To find the length of the texture list
+
+	// printf("cycleLength: %d\n", cycleLength);
```
AngheloAlf: Commeted code

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+	for (currentPtr = textureIndexListOffset; currentPtr < textureIndexListOffset + cycleLength;
+	     currentPtr++)
+	{
+		currentIndex = BitConverter::ToUInt8BE(rawData, currentPtr);
+		// printf("currentPtr: 0x%X\n", currentPtr);
+		// printf("currentIndex: %d\n", currentIndex);
```
AngheloAlf: Same here

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+				        Globals::Instance->inputPath.c_str(), parent->GetName().c_str(),
+				        Seg2Filespace(textureListAddress, parent->baseAddress), texName.c_str());
+			}
+			texturesBodyStr +=
+				StringHelper::Sprintf("    %s,%s\n", texName.c_str(), comment.c_str());
+		}
+
+		texturesBodyStr.pop_back();
+
+		// printf("Declaring texture pointer array\n");
```
AngheloAlf: ditto

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+}
+
+std::string TextureCyclingParams::GetBodySourceCode() const
+{
+	std::string bodyStr =
+		StringHelper::Sprintf("\n    %d, %s, %s,\n", cycleLength,
+	                          parent->GetDeclarationPtrName(textureListAddress).c_str(),
+	                          parent->GetDeclarationPtrName(textureIndexListAddress).c_str());
+
+	// printf("bodyStr = %s\n", bodyStr.c_str());
```
AngheloAlf: Same here

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+{
+	ZResource::ParseRawData();
+
+	TextureAnimationEntry currentEntry;
+	auto rawData = parent->GetRawData();
+	int16_t type;
+
+	for (uint32_t currentPtr = rawDataIndex;; currentPtr += 8)
+	{
+		// printf("Parsing TextureAnimationEntry raw data\n");
```
AngheloAlf: Here too

ZAPD/ZTextureAnimation.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+	for (const auto& entry : entries)
+	{
+		bodyStr += StringHelper::Sprintf("    { %d, %d, %s },\n", entry.segment, entry.type,
+		                                 parent->GetDeclarationPtrName(entry.paramsPtr).c_str());
+	}
+
+	bodyStr.pop_back();
+
+	// printf("bodyStr = %s", bodyStr.c_str());
```
AngheloAlf: Here

AngheloAlf (commented):
Nice work! Just a bunch of minor stuff.
I'm excited for this new extractable resource :eyes:

ZAPD/ZTextureAnimation.cpp
EllipticEllipsis: It would need to be rewritten from a behavioural rather than an extraction perspective, but yes, I assume that MM will eventually have this level of documentation for this, since it's a unique and rather complex set of structs, whether it's written by me or someone a little less prolix.

ZAPD/ZTextureAnimation.cpp
EllipticEllipsis: Left in for debugging purposes, but I suppose unnecessary if we trust the output.

NEstelami: Could enable it if the verbosity level >= VERBOSITY_DEBUG

ZAPD/ZTextureAnimation.cpp
NEstelami: Could the color macros OoT uses be helpful in cases like these?

EllipticEllipsis: Very likely, but I didn't want to import them for just one file's warnings. @AngheloAlf and I working on something project-wide with nicely-formatted warnings like these, so I didn't spend much time on these beyond initial formatting ideas.

AngheloAlf (approved):

NEstelami (approved):

PR #187, opened 2021-09-17
Improve makefiles

EllipticEllipsis:
Made sigificant improvements to the project's three makefiles:
- Move to using build folders to keep everything better-organised.
- Use automatic searching to find files to build rather than mentioning individuals that would need to change if things are moved or extra files are added.
- Integrate the submakes correctly in `clean` and give them their own `.PHONY` rules so that they are always run and are responsible for their own files.
- `export` variables from the main makefile to the submakes, allowing things like the optimisation flags and warnings to be consistent.
- Add an option for using `ld.lld` easily if the user desires, since `LDFLAGS` can't be practically edited from a terminal invocation.
- Remove unused headers.

Also fixed a number of warnings that were missed since -Wextra was not being passed to submakes before, and made the buildinfo into a cpp file, since it contains data.,

### DIFF
--- a/ExporterTest/Makefile
+++ b/ExporterTest/Makefile
@@ -1,18 +1,17 @@
-CXXFLAGS = -Wall -O2 -g -std=c++17
+CXXFLAGS ?= -Wall -Wextra -O2 -g -std=c++17
 OBJ = Main.o TextureExporter.o RoomExporter.o CollisionExporter.o
 LIB = ExporterTest.a
 
-CPPFLAGS-$(MT) += -DCONFIG_MT
-CPPFLAGS += $(CPPFLAGS-y)
 
-.PHONY: all
 all: $(LIB)
 
-.PHONY: clean
 clean:
 	rm -f $(OBJ) $(LIB)
 
-.INTERMEDIATE: $(OBJ)
+format:
+	clang-format-11 -i $(CPP_FILES) $(H_FILES)
+
+.PHONY: all clean format
 
 $(OBJ): Main.cpp TextureExporter.cpp RoomExporter.cpp CollisionExporter.cpp
 
@@ -20,4 +19,4 @@ $(LIB): $(OBJ)
 	$(AR) rcs $@ $^
 
 %.o: %.cpp
-	$(COMPILE.c) $(OUTPUT_OPTION) $(CXXFLAGS) $< -I ./ -I ../ZAPD -I ../ZAPDUtils -I ../lib/tinyxml2
+	$(CXX) $(OUTPUT_OPTION) $(CXXFLAGS) -c $< -I ./ -I ../ZAPD -I ../ZAPDUtils -I ../lib/tinyxml2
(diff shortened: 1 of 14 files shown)


ExporterTest/Makefile
```diff
@@ -1,23 +1,22 @@
...
-.PHONY: all
 all: $(LIB)
 
-.PHONY: clean
 clean:
 	rm -f $(OBJ) $(LIB)
 
-.INTERMEDIATE: $(OBJ)
+format:
+	clang-format-11 -i $(CPP_FILES) $(H_FILES)
```
AngheloAlf, line 20: This makefile doesn't seem to have `CPP_FILES` or `H_FILES` defined. Is that a typo?

Makefile
```diff
@@ -1,8 +1,12 @@
+# use variables in submakes
+export
 OPTIMIZATION_ON ?= 1
 ASAN ?= 0
 DEPRECATION_ON ?= 1
 DEBUG ?= 0
 COPYCHECK_ARGS ?= 
+# Set LLD=1 to use ld.lld as the linker
+LLD ?= 0
```
AngheloAlf, line 9: Could you add this to the readme?

Makefile
```diff
@@ -32,54 +36,76 @@ ifneq ($(DEPRECATION_ON),0)
...
 
-.PHONY: all genbuildinfo copycheck clean rebuild format
+.PHONY: all build/ZAPD/BuildInfo.o copycheck clean rebuild format
 
-%.o: %.cpp
+build/%.o: %.cpp
+	$(CXX) $(CXXFLAGS) $(OPTFLAGS) $(INC) -c $< -o $@
+
+build/ZAPD/Main.o: ZAPD/Main.cpp build/ZAPD/BuildInfo.o
 	$(CXX) $(CXXFLAGS) $(OPTFLAGS) $(INC) -c $< -o $@
```
AngheloAlf: The original reason for this specific rule was to ensure `BuildInfo.h` was updated before compiling `Main.cpp`, but since it now is a standalone `.cpp` file there's no need for this check anymore, so you should be able to remove this rule.

Makefile
```diff
@@ -32,54 +36,76 @@ ifneq ($(DEPRECATION_ON),0)
...
 	$(MAKE) -C lib/libgfxd
 
-ExporterTest/ExporterTest.a:
+.PHONY: ExporterTest
+ExporterTest:
 	$(MAKE) -C ExporterTest
 
-ZAPDUtils/ZAPDUtils.a:
+.PHONY: ZAPDUtils
+ZAPDUtils:
```
AngheloAlf, line 102: Adding those extra `.PHONY` isn't overriding the previous one?

EllipticEllipsis: No, it seems to just treat it like a label for always being out-of-date rather than an actual target. I've seen other makefiles do this, and it seems to work here too. I thought it was worth keeping the submake ones separate from the main file's `.PHONY` targets.

AngheloAlf (approved):

PR #189, opened 2021-09-21
Warning system

EllipticEllipsis:
Following on from fancy warning messages I wrote for TextureAnimation.cpp, and basing the format on the GCC and Clang warnings and errors, @AngheloAlf and I have implemented a set of functions and macros to handle in a consistent way all the errors and warnings ZAPD generates.

Printed warnings and errors are split into three sections:
1. a preamble describing the files and offset involved at the time,
2. a header containing the essential information about the warning type and what caused it
3. a body containing more information that can help fix the warning/error.

They also use colour to stand out in the output of, e.g. `extract_assets.py`:
zapd_warning_example

The functionality and implementation of this system is documented extensively in the file itself and the `README.md`, so I'll be brief here. Some information for the end user is also available via `--help`, although this is still not comprehensive since it's missing flags outside the scope of this PR.
- It is intended to be highly modular, with as much shared functionality as is practical implemented in small shared functions. This makes it simple to add new warning or error categories if desired. 
- Currently it has ordinary (no preamble), PROCESS (processed file printed) and RESOURCE (designed for ZResources, prints the resource name and offset in preamble) categories, depending on where they're used.
- It currently uses a three-level Off/Warn/Err system for warnings and errors, but it would also be possible to include further levels if desired (for example, it we wanted a "fatal error" to stop exectution immediately, but an ordinary error to allow it to continue to provide additional information, or if we wanted a "note" that would print information that is not really a warning), and it would be easy to add this to the current version of the system.
- The default warning information is stored in a `const` map keyed by `std::string`s initially. This and the arguments passed to ZAPD are used to construct a map keyed with the `WarningType` enum holding the information to be used during the rest of exectution, including any overrides of the default levels, 
- Warnings/errors are added to files using macros, which can print the file, line and function for debugging purposes if the verbosity is high enough.
- A high enough verbosity will also print a list of all warning types and their current levels.

When working on this, we have taken the opportunity to attempt to standardise the severity of warnings and errors as well as formatting them to be consistent and easy to spot, hence, for example, any warning relating to invalid XML or image format is now an error.


README.md
```diff
@@ -109,9 +109,49 @@ ZAPD also accepts the following list of extra parameters:
...
+- Warn (print a warning but continue processing)
+- Err (behave like an error, i.e. print and throw an exception to crash ZAPD when occurs)
+
+Each warning type uses one of these by default, but can be modified with flags, similarly to GCC or Clang:
+
+- `-Wfoo` enables warnings of type `foo`
+- `-Wno-foo` disables warnings of type `foo`
+- `-Werror=foo` escalates `foo` to behave like an error
+- `-Weverything` enables all warnings (they may be turned off using `-Wno-` flags afterwards)
+- `-Werror` escalates all enabled warnings to errorsZZ
```
AngheloAlf: Is `errorsZZ` intended?
```suggestion
- `-Werror` escalates all enabled warnings to errors
```

ZAPD/ZTextureAnimation.cpp
```diff
@@ -608,13 +569,20 @@ void ZTextureAnimation::DeclareReferences(const std::string& prefix)
...
+					if (true)
+					{
+						count = 1;
+					}
+					else
+					{
+						[[fallthrough]];
+					case TextureAnimationParamsType::DualScroll:
+						count = 2;
+					}
```
NEstelami: Seems like there's some unreachable code here. Is this some kind of hack/workaround?

EllipticEllipsis: The case label allows you to jump into the else. The idea is that in the SingleScroll case, it sets `count = 1` and falls through to the rest of the scroll-setting code while avoiding setting `count` to `2`, while in the DualScroll case, it jumps into the else, sets it to 2, and carries on to the rest of the code. I thought it was less confusing than setting it to 2 and then setting it to 1 only in the SingleScroll case and using a fallthrough, but maybe not?

NEstelami: I think it would be a good idea to add a comment explaining how that works.

NEstelami (approved):

AngheloAlf (approved):

PR #191, opened 2021-09-23
Add `BaseAddress` check

AngheloAlf:
Adds a small check to prevent the user from inputting a bigger `BaseAddress` than anything pointed by the current file.


EllipticEllipsis (approved):
And this can become a pretty error too once the warning handler's in, of course.

PR #193, opened 2021-09-24
Add `--static` flag

AngheloAlf:
From now on variables won't be declared `static` by default.
A new flag `-s`/`--static` allows to tell ZAPD every asset extracted from a file should be marked static.
A new XML attribute was added to allow marking specific assets in the XML as static. This overrides the behaviour of the global `--static` flag.

This table summarizes if the asset will be marked `static` (✅) or not (❌):
| `Static=""` attribute in XML | Without `--static` flag | With `--static` flag |
|-----------------|-----------------|-----------------|
| `On` | ✅ | ✅ |
| `Global` (default) | ❌ | ✅ |
| `Off` | ❌ | ❌ |

A side effect of this change is now every asset array will be extracted with a explicit length.

Addresses #113


ZAPD/ZSymbol.cpp
```diff
@@ -54,6 +54,9 @@ void ZSymbol::ParseXML(tinyxml2::XMLElement* reader)
 		if (countXml != "")
 			count = StringHelper::StrToL(countXml, 0);
 	}
+
+	staticConf = StaticConfig::Off;
+	// TODO: add a warning/error or something
```
EllipticEllipsis: I think you should either do this, or at least say what it's about...

EllipticEllipsis (approved):

NEstelami (approved):

PR #204, opened 2021-10-19
Plug some large leaks, deprecate `Segment="128"`

### LINKED ISSUE #52: ZAP allocates a lot of memory and resources that are never freed.

EllipticEllipsis:
- Fixed some significant memory leaks in ZCollision, Exporters, ZFile, ZSetMesh.
- Correct ZSkeleton reading out-of-bounds for non-flex skeletons.
- The ZFile leak was caused by non-segmented files not being freed correctly, so I changed ZAPD to assign a default segment (which is the same as the address being VRAM), wrote some more robust checking when reading on the `Segment` attribute, and deprecated `128` as a user input: from now on not declaring the segment will use the default (i.e. VRAM).

This removes 273,662,867 bytes of leaks when extracting OoT; ASan reports no more problems.

Joint work with @AngheloAlf . This should resolve #52 , with the possible exception of the last point.

### DIFF
--- a/ZAPD/ZFile.h
+++ b/ZAPD/ZFile.h
@@ -34,7 +34,7 @@ public:
 	std::map<offset_t, Declaration*> declarations;
 	std::string defines;
 	std::vector<ZResource*> resources;
-	uint32_t segment;
+	int32_t segment = -1;
 	uint32_t baseAddress, rangeStart, rangeEnd;
 	bool isExternalFile = false;
 
--- a/ZAPD/ZFile.h
+++ b/ZAPD/ZFile.h
@@ -34,7 +34,9 @@ public:
 	std::map<offset_t, Declaration*> declarations;
 	std::string defines;
 	std::vector<ZResource*> resources;
-	int32_t segment = -1;
+
+	// Default to using virtual addresses
+	uint32_t segment = 0x80;
 	uint32_t baseAddress, rangeStart, rangeEnd;
 	bool isExternalFile = false;
 
--- a/ExporterTest/Main.cpp
+++ b/ExporterTest/Main.cpp
@@ -76,4 +76,4 @@ static void ImportExporters()
 
 // When ZAPD starts up, it will automatically call the below function, which in turn sets up our
 // exporters.
-REGISTER_EXPORTER(ImportExporters)
+REGISTER_EXPORTER(ImportExporters);
--- a/ZAPD/Globals.cpp
+++ b/ZAPD/Globals.cpp
@@ -23,6 +23,15 @@ Globals::Globals()
 	outputPath = Directory::GetCurrentDirectory();
 }
 
+Globals::~Globals()
+{
+	auto& exporters = GetExporterMap();
+
+	for (auto& it: exporters) {
+		delete it.second;
+	}
+}
+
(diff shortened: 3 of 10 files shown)


ZAPD/ZFile.h
```diff
@@ -34,7 +34,9 @@ class ZFile
 	std::map<offset_t, Declaration*> declarations;
 	std::string defines;
 	std::vector<ZResource*> resources;
-	uint32_t segment;
+
+	// Default to using virtual addresses
```
AngheloAlf, line 38: Shouldn't this say `use` instead of `using`?
```suggestion
	// Default to use virtual addresses
```

fig02: "default to using x" makes sense, you are actively using that default thing

fig02 (approved):

AngheloAlf (approved):

PR #210, opened 2021-12-05
Remove `-Werror` enabled by default and remove `-march`/`-mtune`

AngheloAlf:
Since we can't really check that we are warning-free in every possible compiler anybody would try to build ZAPD with, I decided to disable the use of `-Werror` by default.

I added a `WERROR` variable to the Makefile, passing any non-zero value will pass `-Werror` to the compiler.
Jenkins will pass this variable to the Makefile too, so warnings will still be checked automatically.

I also removed the `-march` and `-mtune` flags because people were having problem with more modern machines.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@ pipeline {
         // Non-parallel ZAPD stage
         stage('Build ZAPD') {
             steps {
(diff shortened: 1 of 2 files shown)


NEstelami (approved):
LGTM

PR #217, opened 2021-12-13
Mac fixes

EllipticEllipsis:
Fix all the warnings that were preventing building on Mac.
- I just removed `-Werror`, since @AngheloAlf 's #210 has a fix for that already.
- I elected to use the PRI macros for printing rather than recast everything. This does introduce inconsistency that we ought to worry about, but the distinction (which causes the errors on Mac and apparently some Windowses), is that these are UB, while the others are already the right type, I believe.
- Yes, the libpng fix for arm64 is nasty. If someone can think of a better way, I'm all ears.


Makefile
```diff
@@ -53,10 +51,23 @@ ifneq ($(LLD),0)
...
+  LDFLAGS += -Wl,-export-dynamic -lstdc++fs
+  EXPORTERS := -Wl,--whole-archive ExporterTest/ExporterTest.a -Wl,--no-whole-archive
+else
+  EXPORTERS := -Wl,-force_load ExporterTest/ExporterTest.a
+  ifeq ($(UNAMEM),arm64)
+    ifeq ($(shell brew list libpng > /dev/null 2>&1; echo $$?),0)
+      LDFLAGS += -L $(shell brew --prefix)/lib
+      INC += -I $(shell brew --prefix)/include
+      else
+      $(error Please install libpng via Homebrew)
```
AngheloAlf, line 70: You could add instructions in the readme on how to install libpng using Homebrew

AngheloAlf (commented):
Overall looks good, just one request

Makefile
EllipticEllipsis: Oh yeah, good point. Done.

AngheloAlf (approved):

NEstelami (approved):

Issue #222, opened 2021-12-24
Glitched Actor lists


Yanis002:
While searching how Impa's working I searched in the scene where the actor is used, and I noticed that the latest alternate header of the spot00 (hyrule field) scene (``spot00_room_0ActorList_001D40``) has a glitched actor list that extends after what should be the end of the array

Yanis002:
(since #266 replaced #224, the fix for this issue, it can be closed now)

Issue #183, opened 2021-09-12
ZAPD should use sizeof


louist103:
Currently ZAPD will extract asset files like this:
```c
Gfx sOtyukaPlatformBottomDL[] = {
    gsSPVertex(&sOtyukaPlatformBottomVtx[0], 4, 0),
    gsSP2Triangles(0, 2, 1, 0, 0, 3, 2, 0),
    gsSPEndDisplayList(),
}; 
```
Assets that are currently in the repo have things like this:
```c
 Gfx sPlatformBottomDList[] = {
    gsSPVertex(sPlatformBottomVerts, ARRAY_COUNT(sPlatformBottomVerts), 0),
    gsSP2Triangles(0, 2, 1, 0, 0, 3, 2, 0),
    gsSPEndDisplayList(),
};
```
Another example of this
```c
CollisionHeader sOtyukaCollision = { 
    { -60, -60, -60 },
    { 60, 0, 60 },
    8,
    sOtyukaCollision_vtx_00002540,
    10,
    sOtyukaCollision_polygons_000024A0,
    sOtyukaCollision_surfaceType_00002490,
    sOtyukaCollision_camDataList_00002488,
    0,
    NULL
}; 
```
Is being extract instead of:
```c
static CollisionHeader sColHeader = {
    { -60, -60, -60 }, { 60, 0, 60 },    ARRAY_COUNT(sVtxList), sVtxList, ARRAY_COUNT(sPolyList),
    sPolyList,         sSurfaceTypeList, sCameraDataList,       0,        NULL,
};
```

Dragorn421:
I think for the vertex buffers it makes more sense to *not* use sizeof / array length since it's quite common to load sub-arrays of the full vertex array into the rsp's vertex buffer

fig02:
agreed, i think it straight up just wont work in many scenarios

AngheloAlf:
I think that's how libgfxd outputs dlists. I don't know if that library allows to change that behaviour.

louist103:
Alright so maybe not for anything GFX related but it should be ok to use for collision right? Or maybe animations of there are any hard codes sizes in there

louist103:
#278

PR #224, opened 2021-12-24
Actorlist fix

AngheloAlf:
Currently ZAPD tries to use as much information as possible to extract the actor lists in scenes instead of using the actual size the command provides. This PR fixes that

Addresses #222


Dragorn421:
This was kind of intended though from what I remember, for example `sutaru_room_0ActorList_000050` (from a test map) has the header saying length 1 but the array obviously has 5 entries

(it may be the only spot where it happens though)

I guess it's not worth making ZAPD extra smart, and for now better to just follow the length from the header.
Maybe eventually the xml could be allowed to override the actor list length, so the sutaru entries don't end up as unaccounted

AngheloAlf:
Yeah you are right.

I added a new resource type so the actorlist can be defined by the user, so for the specific case, this will need to be added to  `sutaru.xml` room_0:
```xml
        <ActorList Name="gSutaruActorList" Count="5" Offset="0x50"/>
```

I just tested this, and it matches with and without specifying that in the XML

Dragorn421:
I just realized having an array that isn't the size specified in the header would be a pain to deal with if considering #183 (basically "use `ARRAY_COUNT`"), if that's to be improved on eventually

One way this could be addressed would be to use
```xml
<ActorList Name="gSutaruUnusedEntriesActorList" Count="4" Offset="0x60"/>
```
instead of `Count="5" Offset="0x50"` to extract the "extra" entries as another array

ZAPD/ZActorList.cpp
```diff
@@ -0,0 +1,194 @@
...
+ActorSpawnEntry::ActorSpawnEntry(const std::vector<uint8_t>& rawData, uint32_t rawDataIndex)
+{
+	actorNum = BitConverter::ToInt16BE(rawData, rawDataIndex + 0);
+	posX = BitConverter::ToInt16BE(rawData, rawDataIndex + 2);
+	posY = BitConverter::ToInt16BE(rawData, rawDataIndex + 4);
+	posZ = BitConverter::ToInt16BE(rawData, rawDataIndex + 6);
+	rotX = BitConverter::ToUInt16BE(rawData, rawDataIndex + 8);
+	rotY = BitConverter::ToUInt16BE(rawData, rawDataIndex + 10);
+	rotZ = BitConverter::ToUInt16BE(rawData, rawDataIndex + 12);
+	initVar = BitConverter::ToInt16BE(rawData, rawDataIndex + 14);
```
louist103: Since this is called `params` in OOT it should probably be that here. Same with `actorNum I think it should be `actorId`

ZAPD/ZActorList.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+	return body;
+}
+
+std::string ActorSpawnEntry::GetSourceTypeName() const
+{
+	return "ActorEntry";
+}
+
+int32_t ActorSpawnEntry::GetRawDataSize() const
```
louist103:
```suggestion
size_t ActorSpawnEntry::GetRawDataSize() const
```

ZAPD/ZActorList.cpp
louist103: Just this and then its good to me.

louist103 (approved):

ZAPD/ZActorList.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+		index++;
+	}
+
+	return declaration;
+}
+
+std::string ZActorList::GetSourceTypeName() const
+{
+	return actors.front().GetSourceTypeName();
```
NEstelami, line 103: Might it make sense to set `GetSourceTypeName()` here to a static function accessed like `return ActorSpawnEntry::GetSourceTypeName()`? It's not an override, and it would ensure that in a scenario where the actor list is somehow empty, it wouldn't try to access a null pointer.

ZAPD/ZActorList.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+ZResourceType ZActorList::GetResourceType() const
+{
+	return ZResourceType::ActorList;
+}
+
+size_t ZActorList::GetRawDataSize() const
+{
+	return actors.size() * actors.front().GetRawDataSize();
```
NEstelami, line 113: Same here.

Issue #227, opened 2021-12-26
Shifting OoT scene headers: `SCENE_CMD_SPECIAL_FILES` should use the `OBJECT_` enum


Dragorn421:
The second argument to `SCENE_CMD_SPECIAL_FILES` is an object id, in vanilla 0 (`OBJECT_INVALID`), 2 (`OBJECT_GAMEPLAY_FIELD_KEEP`) or 3 (`OBJECT_GAMEPLAY_DANGEON_KEEP`)

It could in theory be any object id I think

The issue is ZAPD writes a number instead of the appropriate enum constant, this prevents from shifting the whole object table

--------

Macro definition ([include/z64scene.h](https://github.com/zeldaret/oot/blob/6213fe3bbd0bfcad695640d30d0b0bc8f9faffbe/include/z64scene.h#L472))
```c
#define SCENE_CMD_SPECIAL_FILES(elfMessageFile, keepObjectId) \
    { SCENE_CMD_ID_SPECIAL_FILES, elfMessageFile, CMD_W(keepObjectId) }
```

Relevant struct for the command ([include/z64scene.h](https://github.com/zeldaret/oot/blob/6213fe3bbd0bfcad695640d30d0b0bc8f9faffbe/include/z64scene.h#L72-L76))
```c
typedef struct {
    /* 0x00 */ u8  code;
    /* 0x01 */ u8  cUpElfMsgNum;
    /* 0x04 */ u32 keepObjectId;
} SCmdSpecialFiles;
```

Relevant scene command handler ([src/code/z_scene.c](https://github.com/zeldaret/oot/blob/6213fe3bbd0bfcad695640d30d0b0bc8f9faffbe/src/code/z_scene.c#L234))
```c
void func_8009883C(GlobalContext* globalCtx, SceneCmd* cmd) {
    if (cmd->specialFiles.keepObjectId != 0) {
        globalCtx->objectCtx.subKeepIndex = Object_Spawn(&globalCtx->objectCtx, cmd->specialFiles.keepObjectId);
        gSegments[5] = VIRTUAL_TO_PHYSICAL(globalCtx->objectCtx.status[globalCtx->objectCtx.subKeepIndex].segment);
    }
```

Dragorn421:
I'm not sure if object id 0 where it is used in `SCENE_CMD_SPECIAL_FILES` should be `OBJECT_INVALID` though, 0 is a particular value for the object system,
[...]

PR #228, opened 2021-12-26
Use the object enum in `SCENE_CMD_SPECIAL_FILES`

AngheloAlf:
Addresses #227


NEstelami (approved):

Dragorn421 (approved):
It does solve #227 and the oot object table can be shifted with this :clap: (no immediate crash at least)

PR #230, opened 2021-12-28
Fix `CollisionHeader` extraction

AngheloAlf:
Fixes a bug when trying to extract a `CollisionHeader` which has `CamData` but no `surfaceTypeList`.

To replicate the current bug, use this collision node in the `object_oyu`  XML of MM:

```xml
        <Collision Name="object_oyu_Colheader_000988" Offset="0x988" />
```

Also improved the names of autogenerated variables used by the extracted collision header

### DIFF
--- a/ZAPD/ZCollision.cpp
+++ b/ZAPD/ZCollision.cpp
@@ -77,29 +77,34 @@ void ZCollisionHeader::ParseRawData()
 		polygonTypes.push_back(
 			BitConverter::ToUInt64BE(rawData, polyTypeDefSegmentOffset + (i * 8)));
 
-	if (camDataAddress != 0) {
+	if (camDataAddress != 0)
+	{
 		// Try to guess how many elements the CamDataList array has.
-		// The "guessing algorithm" is basically a "best effort" one and it 
+		// The "guessing algorithm" is basically a "best effort" one and it
 		// is error-prone.
 		// This is based mostly on observation of how CollisionHeader data is
 		// usually ordered. If for some reason the data was in some other funny
 		// order, this would probably break.
 		offset_t upperCameraBoundary = polyTypeDefSegmentOffset;
-		if (upperCameraBoundary == SEGMENTED_NULL) {
+		if (upperCameraBoundary == SEGMENTED_NULL)
+		{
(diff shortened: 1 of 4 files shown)


ZAPD/ZCollision.cpp
```diff
@@ -76,9 +77,25 @@ void ZCollisionHeader::ParseRawData()
 		polygonTypes.push_back(
 			BitConverter::ToUInt64BE(rawData, polyTypeDefSegmentOffset + (i * 8)));
 
-	if (camDataAddress != 0)
+	if (camDataAddress != 0) {
```
EllipticEllipsis:
```suggestion
	if (camDataAddress != SEGMENTED_NULL) {
```
I think?

ZAPD/ZCollision.cpp
```diff
@@ -76,9 +77,25 @@ void ZCollisionHeader::ParseRawData()
...
+		// Try to guess how many elements the CamDataList array has
```
EllipticEllipsis: I think it's worth explaining what the order here is based on.

ZAPD/ZCollision.cpp
```diff
@@ -106,7 +123,7 @@ void ZCollisionHeader::DeclareReferences(const std::string& prefix)
 
 		parent->AddDeclarationArray(
 			waterBoxSegmentOffset, DeclarationAlignment::Align4, 16 * waterBoxes.size(), "WaterBox",
-			StringHelper::Sprintf("%s_waterBoxes_%06X", auxName.c_str(), waterBoxSegmentOffset),
+			StringHelper::Sprintf("%sWaterBoxes", auxName.c_str()),
```
EllipticEllipsis: I'm not sure it's an advantage to drop the offset: might be helpful for debugging?

ZAPD/ZCollision.cpp
AngheloAlf: I added an explanation. Tell me if it is clear enough

EllipticEllipsis: I think probably the clearest explanation would involve giving the most common ordering.

PR #231, opened 2021-12-29
Warn hardcoded pointers

AngheloAlf:
Extends the functionality of  `-Whardcoded-pointer` to every pointer ZAPD has to process, so it will now warn for every raw pointer it will use which has segments 2, 3, 4, 5, or 6, or a VRAM address (0x80XXXXXX).

For other generic / general purposes pointers of other segments that are not catched by that warning, a new warning was introduced: `-Whardcoded-generic-pointer`. It is off by default because how noisy it is.


louist103 (approved):

NEstelami (approved):
I see there are some merge conflicts with this one so I'll be forking off this and making another PR. Good work!

AngheloAlf:
Superseded by #288

PR #234, opened 2022-01-17
Fix use of filesystem in ZTexture

EllipticEllipsis:


fig02:
Merging this in without waiting for the usual approvals-- this is preventing some from building over in MM.

@louist103 fyi because this issue was introduced by your changes, just so you know for next time. Apparently this is only an issue on some setups, which is why jenkins never complained

louist103 (approved):

PR #285, opened 2023-05-01
Additional fixes to include paths on Windows

NEstelami:


Captainkittyca2:
@NEstelami  I tested this and it fixes the issue the Object_link_boy.c file has.

However the inc files themselves still don't generate, and the a.vtx.inc file still generates in the decomp root folder for me. 
image

PR #303, opened 2023-09-14
Update OoT and MM's cutscene names

Yanis002:
This also uses the enums for the types where it fits, currently it's hardcoded but I will use the config files later. Since this is the only thing left to do I was thinking why not opening the PR to get some feedback on things to improve. Also I tried to be accurate to both decomps to keep things understandable


Yanis002:
I removed the hardcoded enums for OoT and instead I'm using a config file, it's an XML file since it makes things easier to manage imo, [here's](https://gist.github.com/Yanis42/b9c0480459b46d1b88d8efe1d341a881) how the file looks like, I'll port these changes to MM later

Edit: [here's](https://gist.github.com/Yanis42/bfc87af02429fb98343c99cdcdbce8f0) MM's XML file (and for both config files just add ``<CutsceneEnumData File="CutsceneEnumData.xml"/>``)

ZAPD/OtherStructs/CutsceneMM_EnumData.h
```diff
@@ -0,0 +1,200 @@
...
+	ENUM_TO_STRING(CS_CMD_ACTOR_CUE_594), ENUM_TO_STRING(CS_CMD_ACTOR_CUE_595),
+	ENUM_TO_STRING(CS_CMD_ACTOR_CUE_596), ENUM_TO_STRING(CS_CMD_ACTOR_CUE_597),
+	ENUM_TO_STRING(CS_CMD_ACTOR_CUE_598), ENUM_TO_STRING(CS_CMD_ACTOR_CUE_599),
+};
+
+#undef ENUM_TO_STRING
+
+// see
+// https://github.com/zeldaret/mm/blob/0c7b90cf97f26483c8b6a98ae099a295f61e72ab/include/z64cutscene.h#L109-118
+const std::string csMMModifySeqTypes[] = {
```
louist103: Unless I missed something these are only accessed with `.c_str()` so they can just be arrays of c strings.

ZAPD/OtherStructs/CutsceneOoT_Commands.cpp
```diff
@@ -0,0 +1,455 @@
...
+}
+
+std::string CutsceneOoTCommand_GenericCameraCmd::GetCommandMacro() const
+{
+	std::string result;
+	std::string listStr;
+
+	if (commandID == (uint32_t)CutsceneOoT_CommandType::CS_CMD_CAM_AT_SPLINE)
+	{
+		listStr = "CS_CAM_AT_SPLINE";
```
louist103, line 183: There can also be c strings.

ZAPD/GameConfig.cpp
```diff
@@ -165,6 +165,54 @@ void GameConfig::ConfigFunc_ExternalFile(const tinyxml2::XMLElement& element)
...
+	     csEnum = csEnum->NextSiblingElement())
+	{
+		for (tinyxml2::XMLElement* item = csEnum->FirstChildElement(); item != nullptr;
+				item = item->NextSiblingElement())
+		{
+			std::string enumKey = csEnum->Attribute("Key");
+			std::string itemID = item->Attribute("ID");
+			uint16_t itemIndex = atoi(item->Attribute("Index"));
+
+			if (!strcmp(enumKey.c_str(), "cmd"))
```
louist103: Either make `enumKey` a c string or use the `==` operator.

ZAPD/GameConfig.h
```diff
@@ -21,6 +21,18 @@ class ExternalFile
 	ExternalFile(fs::path nXmlPath, fs::path nOutPath);
 };
 
+// Stores data from the XML file, the integer is the index (via ATOI) and the string is the value
+class CutsceneData
```
louist103: This should be named something that doesn't make someone think its the actual data.

louist103 (commented):
Looks good so far. I will take a more in depth review when its ready. Here is a few things to get started with.

ZAPD/GameConfig.h
Yanis002: I renamed that to ``CutsceneEnumData``, let me know if you think of a better name

ZAPD/OtherStructs/CutsceneMM_EnumData.h
Yanis002: this file has been deleted since (replaced by xml magic) though I get bad results when using ``char*`` inside the map from ``CutsceneEnumData``, the output is wrong and idk why so I just used c_str instead, let me know if there's a fix or not (I forgot a lot about C++ over the years lol)

Yanis002:
I think this is ready for review now, I did everything I had in mind for cutscenes I believe (also sorry for the big diff)

Yanis002:
new changes since last time: 
- renamed ActorCutscene and old CutsceneEntry
- use enum names (from ``CutsceneEndSfx``) for end sfx inside CutsceneEntry
- renamed the variables to match new CutsceneEntry
- use spawn flag macro when possible
- format the files and renamed "SetActorCutsceneList.h/.cpp" to "SetCutsceneEntryList.h/.cpp"

I didn't change the scene command since it's not changed yet in decomp (I was hesitating, also the config gist has been updated), I don't think I'm gonna do more changes now (for real this time lol)

ZAPD/GameConfig.cpp
```diff
@@ -165,6 +165,80 @@ void GameConfig::ConfigFunc_ExternalFile(const tinyxml2::XMLElement& element)
...
+				cutsceneData.transitionGeneralType[itemIndex] = itemID;
+
+			if (enumKey == "rumbleType")
+				cutsceneData.rumbleType[itemIndex] = itemID;
+
+			if (enumKey == "spawnFlag")
+				cutsceneData.spawnFlag[itemIndex] = itemID;
+
+			if (enumKey == "endSfx")
+				cutsceneData.endSfx[itemIndex] = itemID;
```
NEstelami: Please do `else if` statements here to avoid unnecessary checks.

ZAPD/OtherStructs/CutsceneMM_Commands.cpp
```diff
@@ -1,53 +1,97 @@
...
 	std::string entryFmt = "CS_UNK_DATA(0x%02X, %i, %i, %i)";
+	std::string type = "";
 
 	if (element != csCommandsDescMM.end())
 	{
 		entryFmt = element->second.cmdMacro;
 		entryFmt += element->second.args;
 	}
 
+	if (commandId == CutsceneMM_CommandType::CS_CMD_MISC &&
```
louist103, line 51: `else if` here too.

ZAPD/OtherStructs/Cutscene_Common.cpp
```diff
@@ -0,0 +1,128 @@
...
+/* CutsceneCommand */
+
+CutsceneCommand::CutsceneCommand(const std::vector<uint8_t>& rawData, offset_t rawDataIndex)
+{
+	numEntries = BitConverter::ToUInt32BE(rawData, rawDataIndex + 0);
+}
+
+CutsceneCommand::~CutsceneCommand()
+{
+	for (auto& entry : entries)
```
louist103: This is a vector or pointers so you don't need the reference.

louist103 (changes requested):
Just these two then good for me.

ZAPD/OtherStructs/Cutscene_Common.cpp
Yanis002: I'm not sure to understand this one, this is from the file currently called ``Cutscene_Commands.cpp``, do you meant something like this: ``for (auto entry : entries)``?

louist103: Yes that.

louist103 (approved):
Approved from a ZAPD code perspective. An OOT and MM person may want to verify the docs.

NEstelami (approved):

PR #305, opened 2023-09-20
Update old names to newer ones

Yanis002:
This remove the need to have compatility defines/typedefs in decomps, I made one commit per name update to make reviewing this easier, it targets both games despite the branch having oot in the name, also I ported the xml enum thing from the cutscene PR, I was thinking it would be easier to use this for navi hints

For some reasons I have matching issues with MM, when I remove ``SkelCurveLimbList`` and ``Struct_800A598C`` typedefs it's not matching and I can't figure out why, I feel like it's an issue with object XMLs not having ``Game="MM"`` but idk


Yanis002:
(fixed conflicts)

NEstelami (approved):
LGTM! @louist103 Whenever you get a chance I would appreciate if you could give this a look.

ZAPD/OtherStructs/SkinLimbStructs.h
```diff
@@ -80,17 +80,17 @@ class Struct_800A598C : public ZResource
 	uint16_t unk_0;  // Length of unk_8
 	uint16_t unk_2;  // Length of unk_C
 	uint16_t unk_4;  // 0 or 1 // Used as an index for unk_C
-	segptr_t unk_8;  // Struct_800A57C0*
-	segptr_t unk_C;  // Struct_800A598C_2*
+	segptr_t unk_8;  // SkinVertex*
+	segptr_t unk_C;  // SkinTransformation*
```
louist103: Most of these members could probably be named now.

louist103 (changes requested):
Just this one thing then it looks good to me.

ZAPD/OtherStructs/SkinLimbStructs.h
Yanis002: there's only one thing that wasn't named in decomp yet (``unk_4`` from ``SkinLimbModif``)

I also named the skin limb types thing in the enum since there's something in decomp for that, 4 and 11 values have a signification but other ones will make the segment ``NULL``, that's why the 5th one was removed (and applied formatter as always)

louist103 (approved):

ZAPD/OtherStructs/SkinLimbStructs.h
louist103: I mean name the members in this class.

PR #306, opened 2023-10-08
Use Player Cue Ids enum

Yanis002:
Targets both games, compiles and match for both, same as before to use you will need to add ``<EnumData File="EnumData.xml"/>`` in the ``Config.xml`` file from decomp and create the file ``EnumData.xml`` with the content for the game:
- for OoT: https://gist.github.com/Yanis42/b9c0480459b46d1b88d8efe1d341a881
- for MM: https://gist.github.com/Yanis42/bfc87af02429fb98343c99cdcdbce8f0


louist103:
Works AS is on OOT. Requires the attached patch for MM.
https://gist.github.com/louist103/d2f7c4839d384457cf08872b1965b35e

PR #307, opened 2023-10-20
Use Enum names for Ocarina Actions and Sequence IDs

Yanis002:
forgot to include that in the previous PR, both games match on my side, both gists are updated ([oot](https://gist.github.com/Yanis42/b9c0480459b46d1b88d8efe1d341a881), [mm](https://gist.github.com/Yanis42/bfc87af02429fb98343c99cdcdbce8f0))


louist103:
OOT and MM both OK. MM needs this patch
https://gist.github.com/louist103/d2f7c4839d384457cf08872b1965b35e

PR #319, opened 2024-03-22
Pass more arguments to OoT CS_LIGHT_SETTING

cadmic:
For some unused cutscenes this data might not be 0 so it needs to be preserved for matching


cadmic:
CI should fail for this since the OOT repo needs to be updated first. Is that okay?

PR #322, opened 2024-06-21
ZAPD features to support OOT multiversion assets

cadmic:
See https://github.com/zeldaret/oot/pull/1967


ZAPD/Main.cpp
```diff
@@ -429,6 +435,27 @@ void Arg_CsFloatMode([[maybe_unused]] int& i, [[maybe_unused]] char* argv[])
 	}
 }
 
+uint32_t parse_u32_hex(char* str)
```
louist103: Just change the formatting on this name and its good.

cadmic: Done, thanks

PR #325, opened 2024-11-30
rename CS_BEGIN_CUTSCENE and CS_END

mracsys:
Opening as draft pending concurrence between MM and OoT projects. Ref: https://github.com/zeldaret/oot/pull/2311

Change `CS_BEGIN_CUTSCENE` and `CS_END` macros to `CS_HEADER` and `CS_END_OF_SCRIPT`, respectively. Previous names could be misinterpreted as controlling cutscene behavior when they are only used internally by the cutscene script parser.


mracsys:
No objections from the MM folks.

